PR #6, opened 2017-08-04
Suggested updates to abstract

mmccool:
Some suggested updates to the abstract.   Some of these are cleanups to the English, however others are to clarify the "descriptive vs. prescriptive" purpose of the WoT.   I also explictly mention the building block deliverables.

Let's discuss.   I think the abstract needs a lot of work and since it's the very first thing people will read, it's important to get it right.

### DIFF
--- a/index.html
+++ b/index.html
@@ -38,14 +38,23 @@
   </head>
   <body>
     <section  id="abstract">
-      <p> This document describes architecture on Web of Things which derived from use cases.<br />
-      The architecture consists of building blocks and the function of each block is explained.<br />
-      And it is also described how to map the WoT architecture to the wide variety of real implementation patterns.<br />
-      <br />
-      This document focused on the standardization scope of Web of Things.<br />
-      And other blocks and conditions underlying the Web of Things are described as example implementation and/or requirements such as the description of how to treat constraint devices and security and privacy.<br />
-      Web of Things is distributed platform over the latest Web technology. And Web of Things can wrap the difference among legacy IoT protocols defined by various standardizing organizations and also can provide interoperability between above legacy standards.<br />
-      Utilizing this interoperability, IoT application developers can ship only one wot-compliant application on behalf of shipping each legacy standard-compliant application.</p>
+      <p> This document describes an architecture for the Web of Things.
+      This architecture is derived from a set of use cases and consists of
+      a set of building blocks.
+      Each building block in the architecture is defined and its function explained.
+      The WoT architecture can be mapped onto a variety of real implementation patterns, and this is also described.
+      </p><p>
+      This document is focused on the standardization scope of Web of Things,
+      which consists of three primary building blocks: 
+      Thing Descriptions, a Scripting API, and Binding Templates.
+      Other non-normative architectural blocks and conditions underlying the Web of Things are described as example implementations and/or requirements, such as a description of how to treat constrained devices.
+Non-normative recommendations for security and privacy are also outlined.
+      The Web of Things is a distributed platform using the latest Web technology.
+      The Web of Things also provides for interoperation with other IoT protocols.
+      Primarily, the Web of Things is descriptive, not prescriptive.
+      Utilizing this interoperability, IoT application developers can ship only one wot-compliant application instead of shipping code for multiple standards.
+Conversely, by providing a Thing Description for a legacy device, it is possible to access it from a Web of Things application.
+</p>
     </section>
     <section  id="sotd">
       <p>Through the discussion in WoT IG, This document is reflected the conclusion for the scope of standardization in WoT WG, that is, "Thing Description", "Scripting API" and "Binding Template".<br />
--- a/index.html
+++ b/index.html
@@ -10,7 +10,7 @@
 		  var respecConfig = {
 			  specStatus:   "ED"
 			  ,   editors:      [{name:"Kazuo Kajimoto"}, {name:"Uday Davuluru"}]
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -38,14 +38,23 @@
...
+      a set of building blocks.
+      Each building block in the architecture is defined and its function explained.
+      The WoT architecture can be mapped onto a variety of real implementation patterns, and this is also described.
+      </p><p>
+      This document is focused on the standardization scope of Web of Things,
+      which consists of three primary building blocks: 
+      Thing Descriptions, a Scripting API, and Binding Templates.
+      Other non-normative architectural blocks and conditions underlying the Web of Things are described as example implementations and/or requirements, such as a description of how to treat constrained devices.
+Non-normative recommendations for security and privacy are also outlined.
+      The Web of Things is a distributed platform using the latest Web technology.
```
mkovatsc: I wouldn't call it a platform, as this implies the prescriptive approach.

index.html
```diff
@@ -38,14 +38,23 @@
...
     <section  id="abstract">
-      <p> This document describes architecture on Web of Things which derived from use cases.<br />
-      The architecture consists of building blocks and the function of each block is explained.<br />
-      And it is also described how to map the WoT architecture to the wide variety of real implementation patterns.<br />
-      <br />
-      This document focused on the standardization scope of Web of Things.<br />
-      And other blocks and conditions underlying the Web of Things are described as example implementation and/or requirements such as the description of how to treat constraint devices and security and privacy.<br />
-      Web of Things is distributed platform over the latest Web technology. And Web of Things can wrap the difference among legacy IoT protocols defined by various standardizing organizations and also can provide interoperability between above legacy standards.<br />
-      Utilizing this interoperability, IoT application developers can ship only one wot-compliant application on behalf of shipping each legacy standard-compliant application.</p>
+      <p> This document describes an architecture for the Web of Things.
```
mkovatsc: "an architecture" -- the patterns in this document are quite "meta" and allow for different architectures. I am not sure if this already classifies for an "architectural style". At least we should call it "the overall architecture" (or whatever better suggestion someone might have).

index.html
```diff
@@ -38,14 +38,23 @@
...
+      This architecture is derived from a set of use cases and consists of
+      a set of building blocks.
+      Each building block in the architecture is defined and its function explained.
```
mkovatsc: Currently these explanations are missing. Furthermore, while revising the document so far (in particular the figures), I noticed that it is hard to set "building block" equal to our deliverables. The Binding Templates are, for instance, different from actual protocol bindings. The latter, however, are concrete building blocks. Also the Scripting API is just a part of the overall runtime system for WoT scripts.

My proposal would be to disentangle building blocks and deliverables a bit and explain more what one can implement and what we standardize for that.

index.html
```diff
@@ -38,14 +38,23 @@
...
+      The WoT architecture can be mapped onto a variety of real implementation patterns, and this is also described.
+      </p><p>
+      This document is focused on the standardization scope of Web of Things,
+      which consists of three primary building blocks:
```
mkovatsc: See previous comment.

index.html
```diff
@@ -38,14 +38,23 @@
...
+      This document is focused on the standardization scope of Web of Things,
+      which consists of three primary building blocks: 
+      Thing Descriptions, a Scripting API, and Binding Templates.
+      Other non-normative architectural blocks and conditions underlying the Web of Things are described as example implementations and/or requirements, such as a description of how to treat constrained devices.
+Non-normative recommendations for security and privacy are also outlined.
+      The Web of Things is a distributed platform using the latest Web technology.
+      The Web of Things also provides for interoperation with other IoT protocols.
+      Primarily, the Web of Things is descriptive, not prescriptive.
+      Utilizing this interoperability, IoT application developers can ship only one wot-compliant application instead of shipping code for multiple standards.
+Conversely, by providing a Thing Description for a legacy device, it is possible to access it from a Web of Things application.
```
mkovatsc: Maybe we should start with the "description over prescription" philosophy (i.e., move it to the beginning).

"legacy device" is later used for devices that cannot have a TD or integrators do not want to ahsre, so they are connected through the "proprietary/legacy communication".

We used "existing device" for those which can be wrapped by a TD to become part of WoT.

mmccool:
OK, good feedback.   I admit to having some difficulty deciding what order to put things into, but at least we agree that "Describe don't prescribe" needs to be emphasized.   Also, yeah, "platform" is the wrong word here.   I'll take another stab at it that tries to address these comments...

index.html
mmccool: How about "abstract architecture"?   I'll use that for now while we argue.   I will annotate the issue with a comment.

index.html
mmccool: My current revision adds a bit of explanation to the main deliverables in the abstract and explains how they fit in.   Note that I emphasize the primality of the TD and the secondary (and optional) nature of the Scripting API (and so the WoT runtime).   I've tried to avoid the term "platform" and talk about devices "participating" in the WoT instead (which I think is a better way to thing about it).

mmccool:
OK, I *think* I've addressed the comments with the latest revisions, please take a look.

mmccool:
One other discussion point: I use "device" in several place since using "Thing" (or thing) is a little awkward.   I was also tempted to say "service" instead of "device" in several places.   There are pros and cons to each of these, but in the end just decided to keep it simple and use "device" (lower case, so not a formally defined term, i.e. intentionally vague).

Kazuo-Kajimoto:
merged #6

PR #7, opened 2017-08-04
Security architecture sections

mmccool:
Initial proposals for security sections in architecture document.   Plenty of room for criticism and comment, so fire away.

### DIFF
--- a/index.html
+++ b/index.html
@@ -113,9 +113,42 @@
       </section>
       <section>
         <h2>Security and Privacy</h2>
-        <p> Functional WoT architecture should have the room for providing
-          security and privacy functionalities.<br />
-          In the IoT solutions, once cyber security barrier is hacked, it is more easily led to safety and/or privacy issues than conventional web solutions. That is because hacked IoT devices often treat heat cycle such as central heating systems, physical moving devices such as cars.<br />
+        <p><!-- Requirements -->
+          The functional WoT architecture should provide for
+          best practices in security and privacy. 
+          Generally, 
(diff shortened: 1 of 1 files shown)


mkovatsc:
There are quite a number of unnecessary line breaks. I would suggest a convention where we have one sentence per line. This allows for good diffs and also nicely shows when constructing too long sentences :)

mmccool:
@mkovatsc Re line breaks: actually, I am following Knuth's suggestion for text formatting systems that lines should be broken (in input) at phrasal boundaries, e.g. after commas, etc. not only after periods.   This is to better accomplish the goal you note, to make diffs more meaningful.  I assume you are referring to the input, not the output.   I edited out a bunch of </br> tags that had crept into the HTML probably after conversion from MD but may have missed one, I will check.

mmccool:
A useful reference whose rules I generally try to following in writing.   Most of this applies to the final prose, not the input, but has several useful rules like "Don't start a sentence with a symbol", "Don't make the operator in an equation the only verb in a sentence" and so forth.   

I'm not sure if the convention (break lines at phrasal boundaries) that I'm following is mentioned in here somewhere, it may have been in a LaTeX book I read 20 years ago. 

At any rate, I'm sure we can agree this is not a major issue.  Let's put "tabs vs. spaces" on the list of things to argue about while we are at it ;)

http://www-cs-faculty.stanford.edu/~knuth/papers/cs1193.pdf

mkovatsc:
There was a pragraph close to one word per line, which is kind of another extreme...

PR #16, opened 2017-08-09
Clean up formatting and English

mmccool:
A bunch of changes aimed at cleaning up formatting and English.   These changes DO NOT touch the security and abstract sections in the other two pending pull requests to avoid conflicts.

I have tried to avoid substantive changes in this pull request and aimed only at clarifying the language.  Unfortunately, I think there are some issues around terminology (WoT Servient vs Thing, for instance) that need to be brought into consensus with the rest of the group...

### DIFF
--- a/index.html
+++ b/index.html
@@ -56,8 +56,8 @@
     </section>
     <section>
       <h1>Introduction</h1>
-      <p>This document describes the abstract architecture of the “Web Of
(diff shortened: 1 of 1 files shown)


mmccool:
I went through and commented on the other issues that Matthias raised and I think a number of them are at least partly addressed with this pull request.

PR #29, opened 2017-08-22
Proposed redefinition of "Thing" vs. "Servient"

vcharpenay:
1. modified "links to domain-specific vocabulary" ("Thing Description"). It would be more accurate to say some of the metadata *is expressed* using domain-specific vocabularies.

2. proposed a new definition for "Thing" (main change). Details:
In the current version, the definitions for Thing and Servient are redundant: a Thing is "an entity that that natively implements WoT building blocks (e.g., Protocol Bindings or the Scripting API)." while a Servient is "an entity that implements WoT building blocks".
In the proposed definition, the servient is the software agent and the Thing is the physical entity (with possibly no computational power itself). The Internet of Things was first used to describe the phenomenon that non-computational objects get a digital representation via RFID tags. See Kevin Ashton's article about it: http://www.rfidjournal.com/articles/view?4986. From this perspective, a bottle, a can or any manufactured product can become a "Thing" as long as it has a digital tag (which is not even mandatory if we consider computer vision). To remain aligned with this idea, I think it is important that we define a WoT Thing in such an abstract way that it can be "any physical entity".

### DIFF
--- a/terminology.md
+++ b/terminology.md
@@ -13,6 +13,8 @@ See Thing Description
 ## Thing Description (TD)
 Structured data describing a physical device or software service (WoT Thing) that includes metadata, links to domain-specific vocabulary, a list of offered interactions, the supported protocol bindings for each interaction, and links to related Things.
 
+> Structured data describing a physical entity (WoT Thing) that includes generic metadata, domain-specific metadata, a list of offered interactions, the supported protocol bindings for each interaction, and links to related Things.
+
 ## Interaction Model
 A formal model to semantically and syntactically describe network application interfaces. The Thing Description is built around this model.
 
@@ -40,11 +42,17 @@ The Resource Description Framework (RDF) of the Semantic Web [rdf11-concepts](ht
 ## Thing
 An entity (either a physical device or a software service) whose metadata and interfaces are described by a WoT Thing Description. A Thing can be an existing device (or service) augmented by a Thing Description or an entity that that natively implements WoT building blocks (e.g., Protocol Bindings or the Scripting API).
 
+> A physical entity (either a computational device or a real-word object) described by a WoT Thing Description. Examples of Thing include (but are not limited to) sensors, actuators, building spaces, industrial equipment and manufactured products. A Thing can be interacted with via a WoT Servient.
+
 ## Servient
 An entity that implements WoT building blocks and both consumes Things and exposes one or more Things. Servients usually have multiple Protocol Bindings to enable interaction with different devices. Servients can implement the Scripting API.
 
+> A software agent that implements WoT building blocks and both consumes Thing Descriptions and exposes one or more Thing Description(s). Servients usually have multiple Protocol Bindings to enable interaction with different devices. Servients can implement the Scripting API.
+
 ## WoT Server
 An entity that exposes a network interface consistent with a WoT Thing Description. A Thing is at least a WoT Server. WoT Server is also used to refer to a Servient in server role only.
+
+> An entity that exposes a network interface consistent with a WoT Thing Description. There exists at least one WoT Server for a Thing. WoT Server is also used to refer to a Servient in server role only.
 
 ## WoT Client
 An entity that can connect with ("consume") a network interface described by a WoT Thing Description. WoT Clients usually implement multiple Protocol Bindings. WoT Client is also used to refer to a Servient in client role only.


zolkis:
I would use "Thing" for the abstract entity. Servient is a description of the interaction roles (client + server), and it's not English. So if we were to minimize vocabulary, I'd vote for using Thing and not using Servient in normative (con)text.

vcharpenay:
But I am not saying we should remove one the two terms. On the contrary, they are both very important. They have just different meanings and the current version does not reflect that difference, in my opinion.

mkovatsc:
1. good
2. The point is that a Thing can be both, an existing device/service with TD or something that was created with out building blocks. A Servient is strictly an entity implementing our building blocks and being able to serve multiple Things. Also Things should not be defined as the physical entities, as we use "Things" in our abstraction and more concretely the runtime (see Exposed and Consumed Thing).

The term Servient was newly created to refer to something introduced by WoT.

Kazuo-Kajimoto:
First look, "Thing" is generic which can represent both "physical device" and "logical entity with our building blocks". 
So, current version uses "physical device", "WoT Thing" respectively.
I think current terminology definition is acceptable.

zolkis:
FWIW on OCF they have the _platform_ abstraction representing the hardware that runs the software stack, called _device_ (-->Servient), and the stack runs the _resources_ (--> Things). In fact, platform and device are represented as resources to the outside world.

So in comparison we just seem to miss the platform abstraction, but IMHO we don't strictly need it as an explicit concept, since a Servient may have HW-specific descriptions/Things and multiple Servients may share HW-specific [meta]data. Also, nowadays HW is more and more abstracted away with different levels of virtualization (memory, processor, network, ...).

Therefore I agree with Matthias that we shouldn't tie Things to HW at all, and Servients could just contain the HW-specific [meta]data. From the comments above I see consensus about the principle, we just need to find the right formulations. These are important because our terminology needs to be explained unambiguously to people aware of various standards that operate with [sometimes inconsistently] overlapping terms.

mkovatsc:
I like Zoltan's approach to check if we can name the different entities we have.

Hardware (meaning a physical entity that has processing power, memory, communications interface, power supply, and possibly sensors and actuators) we have usually been calling "device", "gateway", and "cloud server". Here we should stick to English words with "device" being probably the most common. (Sensors, actuators, and other hardware components, we have been calling "local hardware".)

The software stack running on the hardware is either an existing software stack or firmware, or a "Servient".

The abstraction that let's us access hardware from the digital world has been a "Thing" (at least to me). Software stacks can expose one or more Things (e.g., by dividing the hardware into logical components) or by hosting virtual entities).

Each "Things" offers zero or more "Interactions" (which I would set equal to OCF resources). Zero, because there might be a "Thing" with just metadata linking to other Things, for instance, a representation of a room linking to the contained sensor Things.

vcharpenay:
When I say "physical entity", I do not mean computational hardware. Computational hardware is not modeled in WoT, as you mentioned @zolkis, and this seems good to me. However, not only computers are physical in IoT systems. I wrote "Examples of Thing include (but are not limited to) sensors, actuators, building spaces, industrial equipment and manufactured products". Here are three remarks.

A sensor might be modeled as a Thing so that it has geo-coordinates, e.g. in large-scale deployments like cities, but there is no point including hardware details in the information model. What is usually of interest for the application (that is, what needs to be modeled) is the area the sensor covers: a street, a landmark, a parking spot, etc. These can also be Things.

In the OCF terminology, a room in a building is neither a platform nor a device. Yet, in a typical building automation system, it is an element of the information model, which can have properties (like "ambient temperature").

[Section 5.5 of the Architecture document](https://w3c.github.io/wot-architecture/#smart-factory) says "WoT Servient 1 and WoT Servient 2 control factory equipment with legacy communication". Here, the Thing is the "factory equipment" controlled by the servient, in my view.

There is almost nothing to change in the document to be consistent with the definitions I propose (only the first paragraph of the conclusion, maybe?). It is not the goal of WoT to provide a precise definition for Thing, anyway (the term is used only once in the document in fact). Domain-specific semantic models will fill the blanks but our definition should be consistent with such models, which is not at the moment.

mkovatsc:
I was in the process of updating my previous comment; please re-visit.

mmccool:
I need to go check the vocabulary page to see where we are currently but I think the definition of Thing should be consistent with "Something described by a Thing Description" to avoid inconsistencies.  This means it has a network interface, etc.  I also think it would be perfectly fine to have a Thing that is just a Service (eg a software entity running somewhere on the network).  A good example would be a proxy Thing, which may not be directly associated with a physical device, or may abstract it in some way.  Another might be a Thing that provides some generic function, like storage (and some proposed "System Things" will be like this).

There should probably also be a set of type-of subclass relationships between various concepts, for instance, a WoT Servient can be a subclass of a Thing that supports specific functionality (for instance, a WoT runtime, although we can debate this).

It's probably best if I review the wiki and follow up these thoughts there.

mmccool:
To be clear, I disagree that a Thing should be limited to physical devices.  To reiterate, to avoid bizarre linguistic inconsistencies, Things should be "described by a Thing Description".   Certainly concrete physical devices can be an important subclass of things.  But they have to be "smart" and in particular, have network interfaces and able to act as Servers... since that is what Thing Descriptions describe.

vcharpenay:
Yes, sorry, I also was writing my last comment before you posted yours.

Your last definition suits me better, especially the last sentence. Still, I am not quite disposed to say:
 -  that a Thing can be a "service". If a Thing abstracts a room, it sounds strange to speak about some "room service" (no pun intended);
 - that a Thing "abstracts an entity that natively provides a TD". What provides (or exposes) the TD is the Servient, not the Thing. Otherwise, you would end up saying "a Thing exposes an exposed Thing", which is a tautology.

mkovatsc:
@mmccool Have you read my updates here in this thread?

See above, e.g.,
> Thing -- an abstraction of physical entities whose metadata and interfaces are described by a WoT Thing Description. A Thing can abstract an existing device (or service) augmented by a Thing Description or an entity that natively provides a WoT Thing Description by implementing WoT building blocks. Note that physical entities can also be represented by virtual services, e.g., a room whose temperature, brightness, etc. are actually provided by multiple devices located in that room.

mkovatsc:
I will include the updates into the markdown, so we can discuss in the Editors Call.

mkovatsc:
@vcharpenay  please have a look at https://github.com/w3c/wot-architecture/blob/master/terminology.md and confirm that your raised issues were addressed correctly. If so, please close this pull request.

vcharpenay:
OK, this is fine!

Here is a small rephrasing to avoid having "e.g." and then "such as":
> a logical entity (**e.g. a room or building**) that does not directly provide any software stack.

PR #31, opened 2017-08-24
add term: WoT API

mmccool:
The term "WoT API" is used extensively in the security discussion sections.
It might be a good idea to update the WoT Server definition to use this term.
If a different term is deemed better (eg WoT Server API) we will have to update the security sections, so in the interest of time I would suggest deferring such a change until after the FPWD if possible.

### DIFF
--- a/terminology.md
+++ b/terminology.md
@@ -53,6 +53,11 @@ The Resource Description Framework (RDF) of the Semantic Web [rdf11-concepts](ht
 ## WoT Server
 An entity that exposes a network interface consistent with a WoT Thing Description. WoT Server is also used to refer to a Servient in server role only.
 
+## WoT API
+A network interface consistent with a WoT Thing Description exposed by a WoT Server.
+
+> This term is used extensively in the security threat model
+
 ## WoT Client
 An entity that can connect with a network interface described by a WoT Thing Description (i.e., consume a Thing). WoT Clients usually implement multiple Protocol Bindings. WoT Client is also used to refer to a Servient in client role only.
 


terminology.md
```diff
@@ -53,6 +53,11 @@ The Resource Description Framework (RDF) of the Semantic Web [rdf11-concepts](ht
 ## WoT Server
 An entity that exposes a network interface consistent with a WoT Thing Description. WoT Server is also used to refer to a Servient in server role only.
 
+## WoT API
+A network interface consistent with a WoT Thing Description exposed by a WoT Server.
+
```
zolkis: "network interface" has hardware connotations. 
https://en.wikipedia.org/wiki/Network_interface
We need to express the layer WoT is operating on.
We used to have a definition for a "WoT interface" in the IG Best Practices doc. 
http://w3c.github.io/wot/current-practices/wot-practices.html#wot-interface
We'd need to compact that (not very concise) definition.
(thinking on a proposal)

zolkis: What about 

<dt><dfn>WoT Interface</dfn></dt>
<dd>
  Resource-oriented, REST-ful Web interface that allows access to <a>Things</a> exposed over a network using different <a>Protocol Bindings</a>, as defined by <a>Thing Description</a>s.
</dd>

mkovatsc: To me it feels that "REST-ful" is -- yet again -- just used for its buzzwordiness.

The important part was to distinguish "WoT API" from "Scripting API" as many people through those two together. Also some of the Protocol Bindings will not be Web, so it is not a Web interface per se.

---

#### WoT Interface
The network-facing interface of a Thing as defined by its Thing Description. WoT Interfaces are a superset of Web Interfaces, as Protocol Bindings can also include non-Web protocols such as MQTT or ZigBee. While sometimes called "WoT API", it is not to be confused with the Scripting API, which is interfaces with application scripts within a Servient.

mkovatsc: I just notieced, that this requires a tweak to "Interaction Model" and "Interactions", as they describe both, the network-facing interface (WoT Interface) and the application-facing interface (Scripting API).

mkovatsc: Interaction Model definition updated (see https://github.com/w3c/wot-architecture/blob/master/terminology.md)

zolkis: > WoT Interface
The network-facing interface of a Thing as defined by its Thing Description. 

This.

mkovatsc: Meaning taking just the first sentence and move the rest to somewhere in the Architecture document?

zolkis: Yes, I think the first part is concise enough.

zolkis: This is closed now, but the text is not changed.

mkovatsc: https://github.com/w3c/wot-architecture/commit/67ffd835380d53b86260f0240afedda07e7de251

I committed it directly. See entry below RDF in https://github.com/w3c/wot-architecture/blob/master/terminology.md

Or what do you mean?

mmccool:
Zoltan and I discussed this just now and "WoT API" in the security discussion *should* be changed to "WoT Interface".   I will see that that happens in the security pull requests so the security content is consistent with the rest of the document.

PR #38, opened 2017-09-04
Typos, small edits

mmccool:
A bunch of small edits, including typo fixes and grammar cleanups.  The main substantial things I changed:
- Removal of WoT API from terminology.md, since as discussed, "WoT Interface" should be used for the network-facing interface and more specific terms used internally.   I also removed the "sometimes referred to as the WoT API" fuzzyness in the text.
- In the intro there was something about "growing commercial use of the WoT" which I thought was a little weird (chickens and eggs; eggs came first, of course, but were for reptiles...).  I changed it to refer to growing use of web standards in commercial IoT systems.
- Some fuzziness around "System Things" vs "System APIs".  One part of the text implied that System Things must be used for all non-local access, but later the definition of System APIs reverses that.  I change the first mention of System Things to make their use optional, which is consistent with defining System APIs.


Kazuo-Kajimoto:
MMcCool-san,

Thank you very much!

Issue #30, opened 2017-08-23
Consistency check


mkovatsc:
Editors of the other documents should file their concerns, change requests, etc. here in this issue. WoT Architecture Editors will use it as todo list.

danielpeintner:
Besides some more general comments in https://github.com/w3c/wot-architecture/issues/37 I think we might want to adapt the list in 
https://w3c.github.io/wot-architecture/#wot-scripting-api
w.r.t. Scripting API.

I propose to change three Sub APIs bullet list as follows (there is no Discovery  API per se)

* WoT API to discover or retrieve consumed things and create exposed things

* Client API for consuming a thing by means of sending requests to servers in  order to retrieve or update properties, invoke actions, and observe  properties, actions and events. 

* Server API to expose Things over the network and allowing for adding  properties, actions, and events to a Thing.

@zolkis , @knimura, what do you think?

zolkis:
That is correct.

mkovatsc:
"WoT API" would be in conflict with "WoT Interface".

Why is it not called "Discovery API" anymore when it is to discover? I see the part of creating, but saw that as part of the Server API.

mmccool:
I think we should avoid the term WoT API as it's too general and ambiguous (network API or scripting API)?
We were previously using WoT API in the threat modelling document but were referring to the network interface, what we now call the WoT Interface.   We plan to update all the terminology in the security sections to use WoT Interface.

I second the suggestion that we should name scripting sub-APIs after what they are for: Management API, Discovery API, etc.   I also propose that we ONLY use API for things having to do with scripting (eg inside the runtime) and consistently use "Interface" for network interfaces (from outside things).

zolkis:
Right, the generic term "WoT API" should be avoided and use specific terms like WoT Scripting API. We have a `WoT` _object_ as the API entry point in scripting, hence the confusion :).

zolkis:
In Scripting we are calling TD Repository a Thing Directory.
Looking at the [TD Repository term](https://github.com/w3c/wot-architecture/blob/master/terminology.md#td-repository) definition, it appears that Thing Directory would be shorter and more aligned with external terminology.

zolkis:
[...]

PR #39, opened 2017-09-04
Consistency fixes

mkovatsc:
Solves several points raised in #30, especially hub vs gateway.


Kazuo-Kajimoto:
Thank you very much! Mathias-san!

Issue #32, opened 2017-08-24
Security Considerations


mkovatsc:
@mmccool : We still need content for the section "Security Considerations".

That section should clarify that we are primarily describing exsiting IoT platforms and their existing security mechanisms. Thus, in general the security considerations for each IoT platform used apply. Furthermore, we need to summarize what users of WoT building blocks need to consider (e.g., more accessible information might attract more attackers, TDs might leak information, TDs need to checked like any other information consumed on the Web, etc.

Can the security Task Force provide such a text?

mmccool:
Yes, there is some material that ended up in the TD document that should be moved to this document instead as it is of a more general nature.   I plan to create a pull request relative to your restructured version...

mmccool:
Given the timing though we might have to leave those sections empty for now (for the FPWD) but can create PRs to update them ASAP.

mkovatsc:
We need at least an Editor's Note to explain the roadmap.

mkovatsc:
Added Editor's Note.
Up to Security Task Force now...

mmccool:
Thank you.   In a separate issue (#30 Consistency) I responded to an issue that the security section can get very long.  Even to *mention* all the considerations, let alone the threat model, etc. or even to list the *references* will take a lot of space, and will distract people from the meat of the proposal.  There is also an issue that some "cross-cutting" concerns impact more than one architectural component (for instance, TD and Scripting) and we will end up repeating ourselves.

Therefore, I strongly suggest we create a separate deliverable, call it "Web of Things Security Considerations".   We can then (a) summarize the main points in each other deliverable, focusing on issues especially relevant to that deliverable (b) still publish all the details where people can find them, but without distracting people from the main deliverables.

We already arranged for a separate repo, so this will not be difficult to proceed with.

mmccool:
During the Security TF meeting yesterday we discussed the idea of having a separate security deliverable and then just summaries in each of the other documents to both cover all the bases that need to be covered AND keep the complexity of the security sections other deliverables under control.   It was agreed that this was reasonable, with the proviso that any normative points (for example, specific fields in the TD or API calls) should go in the other deliverables and the security document would just be informative (although we do need to publish it in a way we can refer to from the other documents, eg as a W3C note etc).

So that's what we'd like to do.  As discussed before, I think we should create a separate w3c/wot-security repo for this but will use the w3c/wot/security-privacy directory for now.

[...]

PR #40, opened 2017-09-05
Security refinements

mmccool:
Merging various commits and fixes that were in flight from different people.  Revising Ed notes to point to repo (for now, in w3c/wot/security-privacy) where we intend to put some centralized content.

note: I also added anchor tags to some sections and fixed a broken hyperlink using these tags.  These changes do not affect the rendering.

### DIFF
--- a/index.html
+++ b/index.html
@@ -416,23 +416,6 @@
           important to protect IoT systems so that they can not be
           used to launch attacks on other computer systems.
         </p>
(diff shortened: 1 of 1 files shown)


mmccool:
This could be considered relevant to issue #32: 
https://github.com/w3c/wot-architecture/issues/32

PR #54, opened 2018-06-29
Added "Further things to consider" items as per comments at 6/27 main call

ToruKawaguchi:


Kazuo-Kajimoto:
Toru-san's update is reflected

PR #68, opened 2019-01-25
Added new text for requirements

mryuichi:
I added a text of the technical requirements as discussed in the architecture call on Jan. 23rd.
Some modification will be needed according to the comments in that discussion, but nothing is done for this.


mryuichi:
A new text that is described on a simple configuration architecture that matches the requirement I PRed before.

PR #84, opened 2019-02-19
Restructure 13 2

mlagally:
Implementing the changes agreed in architecture call on 7.2.


zolkis:
So many white char changes (coming from editor settings??) that it makes review next to impossible. It's easier to read the resulting text.

mlagally:
@zolkis I reformatted some very long lines to a format that can be handled by Eclipse.
You can try using a HTML diff service, hopefully it is agnostic to these changes.

PR #87, opened 2019-02-26
Issue86 new figures

mlagally:


k-toumura:
I made SVG version of these figures as a trial.  Please feel free to use them.
[arch-high-level-application-device-new.svg](https://github.com/k-toumura/document-for-wot/blob/master/arch-high-level-application-device-new.svg)
[arch-high-level-application-proxy-device-new.svg](https://github.com/k-toumura/document-for-wot/blob/master/arch-high-level-application-proxy-device-new.svg)

Issue #88, opened 2019-02-26
Replace the figure of the Building Blocks


mryuichi:
This issue is a proposal to replace the figure of the Building Blocks to a new one. Because some components of the current Building Blocks have been not well defined yet.  The below is my proposal. 

![images](https://github.com/mryuichi/documents/blob/master/images/arch-new-building-blocks.png?raw=true)

As already discussed, consideration of the use cases and the requirements derived therefrom resulted in noting that there are three types of entities. The figure shows the functional architecture that commonly applies to those three entity types.
[...]

Issue #85, opened 2019-02-20
Hyperlinks to terms


k-toumura:
[Editorial]
In a past teleconference or F2F, there was a discussion about a style of hyperlinks to terminology section.
For example, in [Building Blocks](https://w3c.github.io/wot-architecture/#building-blocks), [Servient Implementation](https://w3c.github.io/wot-architecture/#servient-implementation) and [Security and Privacy Considerations](https://w3c.github.io/wot-architecture/#sec-security-considerations), terms are repeatedly linked to terminology section.  On the other hand, other sections have no links to terminology.

Which style is preferred in this document?

1.   Make all occurrence of technical term be linked to terminology section to clarify their meaning.
2.  Link only from first occurrence of the term for the conciseness.

And, there are lots (55) of broken reference to terminology section.  If needed (especially when we choose an option 1 above),  I'll make a pull request to fix them.   
![screen shot 2019-02-20 at 15 42 36](https://user-images.githubusercontent.com/30313213/53071685-412cdf80-3526-11e9-8592-2606738373dd.png)

k-toumura:
Option 3 (in Feb 20 teleconference)
3. Make all occurrence of technical term be linked to terminology section, but use plain style for these links.  (by @takuki)

For example, [JSON-LD 1.0](https://www.w3.org/TR/json-ld/) uses following style for links to terminology section:
```css
a.internalDFN[data-link-type="dfn"] {
    color: inherit;
    border-bottom: 1px solid #99c;
    text-decoration: none;
}
```

takuki:
@k-toumura, JSON-LD 1.0's way to link to definitions appears very useful and yet non-obtrusive, which I think is exactly what we want in our document as well.

mkovatsc:
+1 for keeping all links but using a more moderate CSS class.

mlagally:
@k-toumura This looks good, can you please create a PR?

mlagally:
PR: https://github.com/w3c/wot-architecture/pull/89

mlagally:
Pr was merged - the issue is resolved.

Issue #83, opened 2019-02-18
Refine Terminology for WoT Runtime


mmccool:
A WoT runtime can include several sub-components to handle communications, scripting, and protocol bindings.  Several aspects of the specification, for example scripting, relate to only one of these components.   This has an impact on how and where we discuss security considerations, among other issues.

Therefore the architecture document should define these sub-components and given them specific names.

mlagally:
@mmccool Which components do you have in mind? Do they have to exist on each devices that carry a WoT runtime?

mkovatsc:
@mmccool It sounds more like you describe the whole Servient. The WoT Runtime should be the component that instantiates Consumed Things and Exposed Things as software objects, manages their state, and generates the TDs. It communicates with other components such as the Binding implementations and application scripts (the WoT Scripting API being one way).

We decided to update [Figure 17](https://w3c.github.io/wot-architecture/images/architecture-implementation.png)  under [Servient Implementation](https://w3c.github.io/wot-architecture/#servient-implementation) with [Figure 18](https://w3c.github.io/wot-architecture/images/arch-simple-conf-wot-servient.png), which needs some more polishing.

mkovatsc:
Also see https://github.com/w3c/wot-architecture/issues/35#issuecomment-467022651

mlagally:
Arch call on 7.3.:

Add the following definition to the terminology section:

The WoT Runtime should be the component that instantiates Consumed Things and Exposed Things as software objects, manages their state, and generates the TDs. It communicates with other components such as the Binding implementations and application scripts (the WoT Scripting API being one way).

mlagally:
Closing the issue to only have one discussion thread.
Continue the discussion in : https://github.com/w3c/wot-architecture/issues/130

PR #89, opened 2019-02-27
non-obtrusive style for dfn links, and some link related corrections.

k-toumura:
Based on Issue #85, I update CSS and other link problems.

Currently, following problems are exist:
- a dangling link in [Servient Implementation](https://w3c.github.io/wot-architecture/#servient-implementation).  I think it will be resolved when Issue #88 and #83 is closed.
- a lot of 'Normative references in informative sections are not allowed.'   The warnings are only displayed when using Firefox.  I think this is a problem of ReSpec.

### DIFF
--- a/index.html
+++ b/index.html
@@ -116,6 +116,13 @@
 		}
 	};
 </script>
+<style type="text/css">
+a[href].internalDFN {
+    color: inherit;
+    border-bottom: 1px solid #99c;
+    text-decoration: none;
+}
+</style>
 </head>
 <body>
 	<section id="abstract">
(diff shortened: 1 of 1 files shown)


mlagally:
@k-toumura unfortunately the CSS fix in your PR did not solve the link problem. Can you please take another look?

k-toumura:
In my browser, links are rendered as shown in below: 

Link appearance in Firefox

(Same as JSON-LD document style; color is as same as ordinal text, underline is thin (1px) solid gray line)

Should we need to more non-obtrusive style?  For example, completely eliminate underline and change background color on mouse-over, as shown in below:

Example of more non-obtrusive style

mlagally:
@k-toumura Thanks, you are right, it looks much better now.
I merged the PR.

PR #92, opened 2019-02-28
Clarify the definition of InteractionAffordance per discussion in TD TF.

takuki:
The TD TF felt the need to clarify the definition of Interaction Affordance, per discussion [here](https://github.com/w3c/wot-thing-description/issues/282#issuecomment-466488572).


mkovatsc:
Before merging this, please confirm with all stakeholders that they want to drop the feature of describing requests using the form mechanism.

Making affordances equal to links does work, but Things will have to implement standardized resources, so adoption to different platforms or existing devices is not possible (but one might be lucky in the cases where it fits by chance).

mkovatsc:
> Metadata of a Thing that links to resources providing a representation of (parts of) the Thing's state to show the possible choices to clients, thereby suggesting how clients may interact with the Thing.

My problems with this are:

* "links to resources": only talks about linking (i.e., how to retrieve information, not act)
* "providing a representation of (parts of) the Thing's state": only covers Properties
* "to show the possible choices": only linking to resources cannot show all the choices, in particular in the context of IoT where it is heavily about actuation

I suggest to revise into this direction, which does limit WoT to link-only controls:

> Metadata of a Thing that combines a well-known application intent with hypermedia controls that describe the possible protocol-level choices to clients, thereby suggesting how clients may interact with the Thing.

vcharpenay:
I recognize the limitations you point out, indeed. Some suggestions:
 - I am not sure I understand the concept of "application intent" here, how about introducing it in the text? We mentioned intents when talking about the Alexa API. The Android architecture also includes an `Intent` class for inter-application communication. But here, what is meant is rather informal and refers to "clients' intent", right? Is it relevant in the definition of Interaction Affordances?
 - I think "hypermedia controls" is worth a little explanation, how about "hypermedia controls (...) to retrieve and modify a Thing's state"?

Here is how I would write it:

> Metadata of a Thing **exposing** hypermedia controls that describe the possible protocol-level choices to clients **to retrieve and modify the Thing's state**, thereby suggesting how clients may interact with the Thing.

Regarding Section 6.4, I would suggest the following changes:

> In this context, an affordance is the description of a hyperlink (e.g. via a link relation type or link target attributes) **or a form (e.g. via field names)** suggesting clients how to act on the linked resource. Drawn from this hypermedia principle, the Web of Things defines Interaction Affordances as metadata of a Thing that **expose hypermedia controls, i.e. a description of the protocol-level choices clients have to retrieve and modify the Thing's state**, thereby suggesting how clients may interact with the Thing. **In WoT, hypermedia controls are given as request templates, which can be seen as forms to fill in and to send to servers.**

What do you think?

k-toumura:
Current definition of [InteractionAffordance in TD document](https://w3c.github.io/wot-thing-description/#interactionaffordance) does not have `links`.  And [Links section in Architecture document](https://w3c.github.io/wot-architecture/#sec-hypermedia-links) says that links are used only for:
 - discovery 
 - to express relations between Things
 - to express relations to other documents on the Web

So, it is not adequate to say that an affordance is description of hyperlink, I think.  What about just saying 'an affordance is a form suggesting clients ...'?

I found the [@mkovatsc's slide in IRTF t2trg](https://github.com/t2trg/hypermedia/blob/master/slides/wot-20181017.pdf). In this slide, @mkovatsc introduces usage of links in WoT Affordance, however, it is currently for 'Educational Purposes' only variant.

takuki:
Hi. Over the weekend, I was also independently thinking the PR needs update to cover actuation better. 

Now looking at @vcharpenay 's suggested text "to retrieve and modify the Thing's state", what would _form_ amount to. Does every [action](url) involve state change? Probably yes, but not all people may not agree. 

I think it is better we focus less on "resources" as @mkovatsc suggests.

vcharpenay:
> Current definition of InteractionAffordance in TD document does not have links.

Note: in the definition I give, I refer to the hypermedia principle _on the Web as a whole_. Only the last two sentences refers to WoT specifically. In this sentence ("Drawn from this hypermedia principle, the Web of Things defines Interaction Affordances as metadata..."), there is no mention of links.

mlagally:
As discussed in the call on 7.3. we will define:

Affordance:
Metadata of a Thing exposing hypermedia controls that describe the possible protocol-level choices to clients, thereby suggesting how clients may interact with the Thing. Examples of affordances include , but are not limited to getting and setting properties, invoking actions, and subscribing to events.

Hypermedia Controls:
hypermedia controls are request templates, which can be seen as forms to fill in and to send to servers.

mlagally:
@takuki I already included the two definitions from above in the "fix terminology"-MR: 
https://github.com/w3c/wot-architecture/pull/115
Please resolve the merge conflict for the other section ouf your PR.

mjkoster:
One refinement I suggest. The event, action, and property definitions with their forms are descriptions of affordances. Invoking an action is not itself an affordance, but a use of an affordance.

In other words, the channel changer knob on a television set is an affordance, the act of changing the channel by using the knob is not.

takuki:
I made changes to this PR to incorporate the decision made during 3/7 telecon, as well as the additional suggestion made by @mjkoster .

PR #93, opened 2019-02-28
Replace the figure of the Building Blocks

mryuichi:
A new figure of the Building Blocks is inserted in the beginning of the section 6.
This building blocks shows common functionalities that devices, applications and proxies must have. It is also based on the description of 5.2 High-level architecture.


takuki:
I think it is important to show a whole functional architecture (the new figure at the front of section 6) first, then the subsequent sections follow to describe each functional building block.

Note that section 7 starts with a figure that again shows all building blocks. My understanding is that it is for describing implementation aspect of each building block, which is different from functional viewpoint that is the focus of section 6.

PR #94, opened 2019-02-28
Add 6.5 Inter connection of servients

mryuichi:
The new building blocks was defined as common functions and structure as a result of the consideration of Use cases and 5.2 high-level architecture. Therefore how to inter-connect among servients should be decribed in the same section of "6. WoT Building Blocks".
On the other hands, the section 8 can be focused to the deployments in the actual situation. For example, "8.5 Devices in a Local Network Controlled from a Cloud" and "8.6 Service-to-Service across multiple Domains" are typical cases in the commercial IoT systems.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1712,15 +1712,15 @@
 				HTTP methods such as PUT and GET are assigned to implement the
 				abstract methods. WoT clients can retrieve and change device
 				settings or device attributes defined in TDs through those abstract
-				methods. In a device servient, TD manager creates an ExposedThing
-				based on TD. TD manager also advertises a TD to other servients, or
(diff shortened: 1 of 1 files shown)


takuki:
> The new building blocks was defined as common functions and structure as a result of the consideration of Use cases and 5.2 high-level architecture. Therefore how to inter-connect among servients should be decribed in the same section of "6. WoT Building Blocks".

I think this PR is addressing how the high-level architecture in section 5.2 gets fleshed out by building blocks. It was originally described as part of the deployment section, but  I think it fits better in the building block section because it is more about how the high-level architecture relates to the building blocks.

PR #95, opened 2019-02-28
Replace the figure of "3.3 use case summary"

mryuichi:
replaced the figure of "use case summary" in section 3.3 and added the text for that.


mlagally:
We discussed this proposal in the architecture call on 1.3. and suggested to change proxy to twin and add the connected car to the image.
Currently the PR has a merge conflict.

mryuichi:
Substituted the figure "use case overview" in https://github.com/w3c/wot-architecture/pull/132.  So this PR has been resolved.

PR #96, opened 2019-02-28
Update 3.2 common patterns

mryuichi:
complete the text and the figures that were missing and modify some description to easily understand the relations with each of pattern in 3.2 and summary in 3.3.


mlagally:
As discussed on 28.2. we would like to use the term "twin" instead of proxy. Will clean up the document later.

mryuichi:
could not remove the conflict, so make PR again later.

Issue #112, opened 2019-03-07
Chapter 4.1.1: needs text for consumer application domain


mlagally:
add language describing scenarios of the consumer domain

ToruKawaguchi:
I made a draft text as PR https://github.com/w3c/wot-architecture/pull/124. Please check.

PR #124, opened 2019-03-10
Added Smart Home as Consumer use case

ToruKawaguchi:
As per https://github.com/w3c/wot-architecture/issues/112


mlagally:
@ToruKawaguchi Thanks for providing this MR. To improve readability and document flow, I reworked the other chapters of the application domain section in:
https://github.com/w3c/wot-architecture/pull/121

Can you please change this MR to also match this form?

ToruKawaguchi:
Updated.

PR #125, opened 2019-03-11
Explainer Updates

mmccool:
This is the result of my editing pass on the WoT-Architecture explainer.   Most of my major edits were at the very start of the document.  The original started with a discussion of "academic publications" which will play into a common criticism of the W3C of being too "academic" and not focused enough on business issues.  Therefore I restructured things a bit to start with a description of the *business* benefits of WoT (reduction of market fragmentation), then the process, then a bit of the history (also cleaning up the incorrect statement about "yearly" workshops).

One other thing I did was try to reduce some of the redundancy where we had listed the three building blocks in three different places using structured lists.  I reduced some of the detail in the second such listing.   I also added back in a missing figure actually showing the building blocks and removed (as discussed) the details around JSON-LD 1.1 (these should be discussed in the TD explainer, however).

One big remaining issue though: I think the list of application domains is too long; but at the same time, we are missing a summary of the "common patterns" and "requirements" that are described in the process, instead going directly from "Use Cases" to "What is being Standardized".   I think we need to (a) cut down the size of the Use Cases section; it is enough to say we considered use cases in the "following areas" and then list the section headers in the current draft in simple paragraph.  The "common patterns" can use the summary picture showing the network configuration that we have been working on; then "requirements" can be a paragraph or two...

I can take a stab at filling in these sections after we discuss for a bit more here what I've done so far.


mmccool:
Argh... @mlagally just merged a bunch of changes that conflict in major ways with this PR so a merge just gives a big mess.   Michael, I will leave it up to you to take a look at my version and see what you can work out.

mlagally:
@mmccool This is very difficult to review against the new version of the explainer, since the github diff shows the previous version of the explainer. Is there anything you can do about it, e.g. update the PR?

k-toumura:
GitHub has a function to compare any two commits or forks.
https://help.github.com/en/articles/comparing-commits-across-time

For example, following url compares @mmcool's fork and w3c:master.
https://github.com/w3c/wot-architecture/compare/mmccool:explainer...master

mmccool:
Yes, like I said, there are enough differences that github merge just makes a mess.   I'm probably going to abandon this PR, but checkout your new version and try to work my edits back into that basis.    So, don't try to review this PR, it's hopeless.

mmccool:
Anyway, starting work on this now.   If anyone is also working on it in parallel please let me know... I'll try to complete my edits in a timely manner (eg within a few hours from now).

mmccool:
OK, now ready for review.   Since I created a "Basis" document, to review rather than trying to use "git diff", you should diff the Basis against the main document.  To make this easier for you I also checked in a "Diff" document.  The diff is not perfect, but...

I'm going to leave this for now and turn my attention to the TD Explainer.  I should note there are still a number of empty sections.  In particular I think we need to summarize the "Common Patterns" and "Requirements" in appropriate sections as marked.

mmccool:
Found a bunch of small problems upon re-reading it so there are a number of small commits above.   there are probably some problems still, however, I'm going to force myself to stop looking at it now ;).  The Diff does reflect the state of the master after each commit (including the last one).

mlagally:
@mmccool Many thanks for your work on improving the explainer. 

To keep things simple, I will merge it now. We should get rid of the other files (baseline etc.) soon.

I'm not sure if we get additional contributions before our next architecture call, so we should consider dropping the empty sections.

One more comment: There's a lot of content in the document now, at some point I believe we have to pick and chose, since we we otherwise would be just replicating major parts of the architecture document.

mmccool:
Regarding extra context we could drop, I think the use cases have more detail than necessary.   I would replace them with a simple paragraph that just lists the application domains considered and say "a large number of concrete use cases were considered in these application domains".   Another point where we could drop content is the whole "Architectural Aspect" thing.   I think it is useful for the main text but maybe we should just dive into the descriptions of the Building Blocks directly.

At any rate, yes, dropping the missing sections would be ok at this point.

Let me create another PR to try and clean up some of the above... we can discuss in the main call and try to get closure then.

mmccool:
See https://github.com/w3c/wot-architecture/pull/129

mmccool:
I decided to keep the "architectural aspects" section... If you still think the explainer is too long we could probably still write it out, but I think it makes a good transition.

PR #138, opened 2019-03-16
Remove TD-specific content from explainer

mmccool:
Ready to Merge

TD-specific content moved to the TD explainer is deleted from the Architecture explainer in this PR to avoid redundancy.
1. Remove JSON-LD 1.0 TD design alternative


mmccool:
Ready to merge.   In the end I only deleted one paragraph (and moved it over to the TD explainer).

Issue #130, opened 2019-03-13
Clarify WoT Runtime vs other known runtimes


zolkis:
WoT uses the word "Runtime" in a less than well defined manner.
We should explore and use terminology from other runtime environments.
Since the current reference implementation is TypeScript, the first runtime environment to check is ECMAScript. In particular, the following notions should be mapped/discussed:
- [execution context](https://tc39.github.io/ecma262/#sec-executable-code-and-execution-contexts)
- [agent](https://tc39.github.io/ecma262/#sec-agents)
- [built-in objects](https://tc39.github.io/ecma262/#sec-ecmascript-standard-built-in-objects), e.g. the [global object](https://tc39.github.io/ecma262/#sec-global-object)
- [data types/structured data](https://tc39.github.io/ecma262/#sec-global-object) in conformance with TD definitions
- [control abstractions](https://tc39.github.io/ecma262/#sec-control-abstraction-objects), e.g. Promise
- the HTTP operations/client library used (e.g. fetch())
- and possibly other things.

I also propose merging Section 8 (some Servient implementation details, not much text, repeated sentences) into Section 7 (Building Blocks), split where it applies. IMHO in general the Architecture doc should be more about the building blocks rather than specific implementation issues, but when the latter arise, could be encapsulated in the building blocks text. Maybe Implementation Guidelines are also needed, but then the current Section 8 is not exactly that.
(this might be split in another issue).

zolkis:
Related to #83

mlagally:
As discussed in arch call on 14.3.:

The current WoT runtime definition is language independent by intention:
A runtime system for application scripts with the WoT Scripting API. Implementing a WoT Runtime is optional for Servients. The WoT Runtime should be the component that instantiates Consumed Things and Exposed Things as software objects, manages their state, and generates the TDs. It communicates with other components such as the Binding implementations and application scripts (the WoT Scripting API being one way).

We should not limit it to a specific language / execution environment.
Would a different term help to resolve your concerns (e.g. WoT Runtime Context)?

The split between chapter 7 and chapter 8 is intentional to separate Servient implementation from building blocks.

zolkis:
Right, the WoT Runtime is language independent, but we need to operate with generic runtime concepts, such as the ones enumerated above. This is important e.g. for security considerations. The WoT Runtime imposes a set of constraints on top of generic runtimes. It is important to specify what are those.

[...]

PR #143, opened 2019-03-20
Runtime/Building Block/Servient Implementation Refinements

mmccool:
An attempt to address https://github.com/w3c/wot-architecture/issues/130 and resolve some ambiguities around whether or not the WoT Runtime is a required component, this PR includes revisions to chapter 6 and 7 that are based on the assumption that the WoT Runtime can be used WITHOUT the WoT Scripting API.  As such, it includes an update the definition of WoT Runtime in the Terminology section as well.

This PR also adds security content to the WoT Building Blocks and in a few other places.


mmccool:
NOTE: based on the discussion in the main call, will either complete this PR (and layer a new abstraction on top of the WoT Runtime, making the WoT Runtime specfic and required for the Scripting API) OR create another PR and make the WoT Runtime required for a WoT Thing, but make the WoT Scripting API *optional* for the WoT Runtime.

index.html
```diff
@@ -1762,79 +1769,103 @@ <h1>WoT Building Blocks</h1>
...
-			It is also possible, as shown in <a
-				href="#arch-thing-without-scripting"></a>, to use the WoT Thing
-			Description to described a device or service that does not use the
-			WoT Scripting API. There are two cases where this may be important:
-			pre-existing "brownfield" devices for which we are providing a Thing
-			Description in order to allow them to interoperate with other
-			elements of a WoT system; and devices which need to use a programming
-			language not supported by the WoT Scripting API.
+                <p>It is however possible to implement a WoT Thing <em>without</em> using the WoT Runtime
+                and WoT Scripting API as shown in <a href="#arch-thing-without-scripting"></a>.
```
zolkis: Somehow we need to make clear that the WoT Runtime is larger concept than the Scripting Runtime used in the implementation of the Scripting API.

index.html
```diff
@@ -1910,17 +1941,18 @@ <h3>WoT Binding Templates</h3>
...
 				(Properties, Actions, and Events). It also allows scripts to specify
-				and implement Interaction Affordances and expose Things.</p>
+				and implement Interaction Affordances and expose Things (that is,
+                                define and publish a Thing Description).</p>
 			<p>
 				The <a>Scripting API</a> is an optional "convenience" building block
-				in WoT and it is typically used in network nodes that are able to
-				host a WoT Runtime, for instance gateways or more powerful IoT
-				devices.
+				in WoT.  It is used in combination with a WoT Runtime
```
zolkis: I wonder if we are making a constraint here. So far I understood that a WoT Runtime can exist without a scripting API implementation, just encapsulating the WoT network interface and the environment.

index.html
mmccool: Yes, I think your comments were a little behind my edits.

mmccool:
Ready to merge.  It only includes edits up to "Servient Interconnection", however.   But that should be OK since those sections are about external connections and this set of changes focuses on the internal architecture of Things and Servients.

mlagally:
@mmccool Some initial review comments:
* The PR breaks the section structure of the chapters following chapter 7.
* Figure 24 should not contain the ThingDirectory, this will confuse the reader

mlagally:
Agreed an arch call on 21.3.:

WoT Runtime: A runtime system that maintains an execution environment for applications and is able to expose and optionally consume a TD, offers a protocol binding to one or more protocols, and maintains private security metadata.

mlagally:
Scripting Runtime:
A runtime system that is used for implementing an API in a scripting language (e.g. using ECMAScript, LUA, Python etc.)

Issue #149, opened 2019-03-21
Describe WoT Server and WoT Client at the beginning of section "6. WoT Architecture"


ashimura:
As I mentioned during the WoT Architecture call the other day, there was some discussion on the WoT Architecture during the Scripting call as well and we (=myself, Zoltan, Mizushima-san, etc.) started to think it would be easier to understand the concept of "Servient" and "WoT Architecture" if we explained what "WoT server" and "WoT clinet" at the top of section "[6. WoT Architecture](https://w3c.github.io/wot-architecture/#sec-wot-architecture)".

The [Figure 14](https://w3c.github.io/wot-architecture/images/arch-use-case-overview.png) already mentions server and client, so it would be quite natural for us to continue the description based on "server" and "client" as the basis of defining the "Servient".

For example, section "[6. High-level architecture](https://w3c.github.io/wot-architecture/#sec-WoT-servient-architecture-high-level)" starts with the simplest structure with an application and a device, and then descibes the 3-piece structure with an application, a proxy and a device, but probably we should mention the basic mechanism (1) a WoT server within a device exposes the capability of the device using a Thing Description instance attached to the device itself and (2) a WoT client within an application consumes the capability of the device using the exposed Thing Description instance.

mlagally:
call on 21.3.:
update figure to replace application with client and device with server.
Align text to explain.
Need to define a servient here, explain that a proxy has client and server role.
Remove figure 17.
Update figure 18: call middle box a Servient with device, server and a "proxy" label on top.

mlagally:
PR: https://github.com/w3c/wot-architecture/pull/158

PR #155, opened 2019-03-21
Update figures in High-Level Architecture section

k-toumura:
For Issue #149.


mlagally:
@k-toumura Many thanks for your updates. In today's call we wanted to get rid of the "device" and "application" terms at all to keep the diagrams more generic, i.e. make them applicable also to cloud services etc. Can you please update and remove these terms? We only wanted to keep the term "proxy" in the second figure.

PR #157, opened 2019-03-21
Correct and add Scripting related content. Fix some of the whitespace…

zolkis:
… problems.

Signed-off-by: Zoltan Kis <zoltan.kis@intel.com>


mmccool:
The changes say these objects are created by the Scripting API, which does not align with the use of these objects in the alternative native implementation given in section 8.10.1 (which does not use the Scripting API).  I would prefer reverting back to "Server API" and "Client API" to avoid a dependence on the "Scripting API" and then treat the Server API and Client API as a set of abstractions that the Scripting API implements, but that can also be implemented (conceptually) in other languages.   This probably has some impact on how the Scripting API spec should be structured (eg as a set of abstractions for the Server and Client APIs, followed by a binding to ECMAScript).   One thing we might want to discuss is whether we need the terms "ClientAPI" and "ServerAPI" at all.   As far as I am concerned "ExposedThing" and "ConsumedThing" ARE the abstractions (they are just object-oriented APIs for the Server API and Client API functionality, respectively).

mmccool:
Maybe we should say these objects represent client and server APIs (no caps, remove special term) that are created by the WoT Runtime (not the scripting API) to represent interfaces to Things via the Protocol Bindings.   I think that aligns with everything else.   We could add an additional sentence somewhere that says that when the Scripting API is used, these objects are bound to objects in the scripting language, if that makes sense.

PR #172, opened 2019-03-23
TAG submission documents

mmccool:
Ready for review/merging

- Small updates to the TAG submission request, but then a big new document, the answers to the security and privacy questionnaire.   Some of these answers are related to the WoT Scripting and WoT Thing Description... I think it might be easier for everyone just to have one document for ALL WoT documents rather than trying to do one for each.
- I may also update some of the Security and Privacy considerations.  
     - In particular I think it might also be useful to have a "Scope" section somewhere.  In particular, I say 
       certain things are out of scope, like provisioning of keys, but the discussion of LCM in the architecture 
       doc seems to imply otherwise.   It would be safer to limit the scope of the WoT Architecture to the 
       "operational" phase and NOT make it a responsibility of the WoT Runtime.
     - Not in this PR, though...
- ~~I accidentally included a commit that expands all remaining tabs in index.html to 8 spaces.  However, that commit changes NO actual content.  I think all the tabs that were expanded were actually just in the HTML head (scripts and references) anyhow.~~ (REVERTED; should deal with separately)

### DIFF
--- a/index.html
+++ b/index.html
@@ -7,146 +7,146 @@
 <script class="remove"
     src="https://www.w3.org/Tools/respec/respec-w3c-common"></script>
 <script class="remove">
-	var respecConfig = {
-		specStatus : "ED",
-		processVersion : 2017,
-		shortName : "wot-architecture",
-		copyrightStart : 2017,
-		wg : "Web of Things Working Group",
-		wgURI : "https://www.w3.org/WoT/WG/",
-		wgPublicList : "public-wot-wg",
-		edDraftURI : "https://w3c.github.io/wot-architecture/",
-		githubAPI : "https://api.github.com/repos/w3c/wot-architecture",
-		issueBase : "https://www.github.com/w3c/wot-architecture/issues",
-		editors : [ {
-			name : "Matthias Kovatsch",
-			w3cid : "75998",
-			company : "Huawei",
-			companyURL : "https://www.huawei.com/"
-		}, {
-			name : "Kazuo Kajimoto",
-			w3cid : "74112",
-			company : "Panasonic Corp.",
-			companyURL : "https://www.panasonic.com/"
-		}, {
-			name : "Ryuichi Matsukura",
-			w3cid : "64284",
-			company : "Fujitsu Ltd.",
-			companyURL : "https://www.fujitsu.com/"
-		}, {
-			name : "Michael Lagally",
-			w3cid : "47166",
-			company : "Oracle Corp.",
-			companyURL : "https://www.oracle.com/"
-		}, {
-			name : "Toru Kawaguchi",
-			w3cid : "79307",
-			company : "Panasonic Corp.",
-			companyURL : "https://www.panasonic.com/"
-		} ],
-		otherLinks : [
-				{
-					key : "Contributors",
-					data : [ {
-						value : "In the GitHub repository",
-						href : "https://github.com/w3c/wot-architecture/graphs/contributors"
-					} ]
-				}, {
-					key : "Repository",
-					data : [ {
-						value : "We are on GitHub",
-						href : "https://github.com/w3c/wot-architecture/"
-					}, {
-						value : "File a bug",
-						href : "https://github.com/w3c/wot-architecture/issues"
-					} ]
-				} ],
-		localBiblio : {
-			"CoRE-RD" : {
-				href : "https://tools.ietf.org/html/draft-ietf-core-resource-directory-11",
-				title : "CoRE Resource Directory",
-				status : "Internet-Draft",
-				publisher : "IETF",
-				date : "03 July 2017"
-			},
-			"WOT-SECURITY-CONSIDERATIONS" : {
-				href : "https://www.w3.org/TR/wot-security/",
-				title : "Web of Things Security and Privacy Considerations",
-				publisher : "W3C"
-			},
-			"WOT-SECURITY-BEST-PRACTICES" : {
-				href : "https://www.w3.org/TR/wot-security-best-practices/",
-				title : "Web of Things Security Best Practices",
-				publisher : "W3C"
-			},
-			"WOT-SECURITY-TESTING-PLAN" : {
-				href : "https://www.w3.org/TR/wot-security-testing-plan/",
-				title : "Web of Things Security Best Practices",
-				publisher : "W3C"
-			},
-			"IEC-FOTF" : {
-				href : "https://www.iec.ch/whitepaper/pdf/iecWP-futurefactory-LR-en.pdf",
-				title : "Factory of the future",
-				publisher : "IEC",
-				date : "October 2015"
-			},
-			"iot-schema-org" : {
-				href : "https://iot.schema.org/",
-				title : "iot.schema.org"
-			},
-			"REST" : {
-				title : "REST: Architectural Styles and the Design of Network-based Software Architectures",
-				author : "Roy Thomas Fielding",
-				status : "PhD thesis",
-				publisher : "University of California, Irvine",
-				date : "2000"
-			},
-			"SAREF" : {
-				href : "https://sites.google.com/site/smartappliancesproject/ontologies/reference-ontology",
-				title : "Smart Appliances REFerence (SAREF) ontology",
-				publisher : "ETSI",
-				date : "November 2015"
-			},
-			"SSN" : {
-				href : "https://www.w3.org/TR/vocab-ssn/",
-				title : "Semantic Sensor Ontology",
-				publisher : "W3C",
-				date : "October 2017"
-			},
-			"HMI" : {
-				title : "The Encyclopedia of Human-Computer Interaction, 2nd Ed",
-				author : "Mads Soegaard, Rikke Friis Dam",
-				publisher : "Interaction Design Foundation",
-				date : "2013"
-			},
-			"NORMAN" : {
-				title : "The Psychology of Everyday Things",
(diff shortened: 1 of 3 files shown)


proposals/TAG review request.md
```diff
@@ -11,18 +11,25 @@ I'm requesting a TAG review of:
...
   - [X] I have reviewed the TAG's [API Design Principles](https://w3ctag.github.io/design-principles/)
 
 You should also know that...
 
-[please tell us anything you think is relevant to this review]
+- The [WoT Security and Privacy Considerations](https://github.com/w3c/wot-security/) document provides extra discussion of security with an IoT focus.
+- The related [WoT Thing Description](https://github.com/w3c/wot-thing-description) document is being submitted for TAG review at the same time.
+    - Tests of implementations will be associated with this document
+- The [WoT Binding Templates](https://github.com/w3c/wot-binding-templates) non-REC track informative document describes an optional WoT building block.
+- The [WoT Scripting API](https://github.com/w3c/wot-binding-templates) non-REC track informative document describes an optional WoT building block.
```
mlagally: is there a better term for "non-REC track informative document"? 
e.g. informative WG note?

proposals/security_and_privacy.md
```diff
@@ -0,0 +1,277 @@
...
+This is not a browser-oriented specification although
+WoT Things can be accessed through browsers as if they
+were web services. 
+However, no new state is introduced or required in browsers beyond
+current mechanisms (cookies, etc).
+
+WoT Things can also act as clients, however, accessing
+web services and other WoT Things acting as servers.
+It is also possible that the same device could instantiate
+multiple WoT Runtimes.  However, there is noi explicit mechanism
```
mlagally: typo: noi

proposals/security_and_privacy.md
```diff
@@ -0,0 +1,277 @@
...
+Yes.  It enables WoT Things to run applications based
+on ECMAScript, using a constrained API, and with suitable
+restrictions, such as no direct access to private
+security data.
+
+However, a mechanism for loading new scripts is not defined
+and is currently considered to be out of scope.
+Our architecture assumes the scripts running inside a WoT Thing
+are trusted implementations provided by the manufacturer
+of the device and installed using a suitable OTA update
```
mlagally: I would put OTA in brackets, since you could have other mechanisms too

proposals/security_and_privacy.md
```diff
@@ -0,0 +1,277 @@
...
+mechanism; but this update mechanism is not defined in our
+architecture and is also currently considered to be out of scope.
+
+### Does this specification allow an origin access to a user’s location?
+
+As noted above in the PII response, in theory a device
```
mlagally: Assigning a location to a device is common practice in industrial use cases. Suggest to remove "in theory".

proposals/security_and_privacy.md
```diff
@@ -0,0 +1,277 @@
...
+on behalf of a WoT Client are those that it chooses
+to support via its exposed interactions (network API).
+
+### Does this specification allow an origin access to other devices? 
+
+Yes, as discussed above.  A WoT Thing can act as a gateway to 
+other WoT Things or to other non-IP protocols.
+
+The situation is unavoidably more complex than the client-server
+architecture of broswers.  A better analogy than browsers
```
mlagally: typo: browser

proposals/security_and_privacy.md
```diff
@@ -0,0 +1,277 @@
...
+included in data or metadata provided by WoT Things.
+The only operations that a WoT Server will perform
+on behalf of a WoT Client are those that it chooses
+to support via its exposed interactions (network API).
+
+### Does this specification allow an origin access to other devices? 
+
+Yes, as discussed above.  A WoT Thing can act as a gateway to 
+other WoT Things or to other non-IP protocols.
+
```
mlagally: Access to other devices is one of the primary use cases of WoT.

proposals/security_and_privacy.md
```diff
@@ -0,0 +1,277 @@
...
+This concept is not applicable to the WoT context.
+
+### How should this specification work in the context of a user agent’s "incognito" mode?
+
+This concept is not applicable to the WoT context, as there is no user agent.
+
+### Does this specification persist data to a user’s local device? 
+
+Potentially, as a WoT Thing may include interactions that
+support POST requests.  IoT devices can act like web servers that
```
mlagally: The question is what is meant with "the user's local device". I would interpret it as the device, from which the user accesses and controls an IoT device. 
The specification does not persist data, it is completely under control of the user's device (and also of the IoT device) whether it persists some state changes.

mlagally:
The new document is great work! 
There is the tab vs. space problem which creates a noisy diff, which we can resolve in a future commit.
Github allows to ignore whitespace changes.

index.html
```diff
@@ -7,146 +7,146 @@
...
-				title : "The Constrained RESTful Application Language (CoRAL)",
-				author : "Klaus Hartke",
-				publisher : "IETF",
-				status : "Internet-Draft",
-				date : "2019"
-			}
-		}
-	};
+        var respecConfig = {
+                specStatus : "ED",
```
zolkis: Nowadays tabs are expanded to 2 or 4 spaces in these documents, but it doesn't matter.

proposals/TAG review request.md
mmccool: Sure, I can change that so it's positive rather than negative.  BTW at some point I will revise the TD note as well and will make them consistent.

proposals/security_and_privacy.md
mmccool: What, noi is noi a word?   Anyhow... I will fix.   I guess running a spell checker over this document would not hurt...

proposals/security_and_privacy.md
mmccool: True, it was just meant to be an example.   Maybe I should just say "for example...".   BTW I plan to update the security considerations section of the main document to say something similar.  It occurs to me we did not discuss scope, operational phase, etc. in the main document.   I probably should also talk about cross-site issues, single origin, shared data, Zero-Trust, etc.  Most of these are irrelevant to IoT, or rather, a different approach needs to be taken to software distribution.   This is why I was reluctant to include scripts in the TD.  If we did include them avoiding cross-site scripting hacks would be a real nightmare given the more general connections used in IoT architectures.

proposals/security_and_privacy.md
mmccool: Hmm... let me think about it.   Maybe I should differentiate different use cases here: industrial, personal, and medical, which have different requirements.   Probably should include those different views in the main doc too.

proposals/security_and_privacy.md
mmccool: Yep, I probably should say something strong here.   I will steal your sentence.

proposals/security_and_privacy.md
mmccool: Later on I talk about this more.   In the "normal" web, the device is a client, but downloaded scripts can persist state, which causes various issues (for example, tracking).   I was thinking here about the case of a personally owned device which is a server, which some other device posts data to and then retrieves it later.   There is nothing stopping someone from creating a WoT device that acts like a blackboard (for example, to allow messages to be posted to a watch display) and allows set and get of strings, for instance.   Such a WoT Thing would be privacy risk since a tracking ID could be posted to it...

index.html
mmccool: Yeah, I may do some cleanup.   At any rate, I got rid of the tabs, and intend to keep them out of the document.  VI set up now; sorry, some of the tabs might have been my fault; new machine, forgot to set up VI "properly".

mmccool: This was causing conflicts so I just reverted this file to the one in the master repo.   We can deal with the tab issue later...

proposals/TAG review request.md
mmccool: Changed in both occurrences to "informative WG note describing an optional WoT building block."

proposals/security_and_privacy.md
mmccool: I decided to just remove OTA, it's not really relevant whether the update is OTA or not.

proposals/security_and_privacy.md
mmccool: Hmm... note I was talking about "the user's location" not "the device's location".   The "in theory" part is because in some cases you can infer the user's location from the device's location.  But in some cases that may be intentional.  The privacy hazard is that you don't want *unintentional* disclosure of user location.

mmccool: I deleted the "in theory", and added a sentence stating that is in fact common in institutional use cases.

proposals/security_and_privacy.md
mmccool: OK, I updated this to talk about both cases: the device is a server, and the device is a client.  Persistent store-and-return data storage is possible in both cases.  Mitigation in both cases is to use access controls and establish trust before storing and returning data.

proposals/security_and_privacy.md
mmccool: Fixed up, I think.  Expanded explanation of "Zero-Trust"; this should go into the main document, too.   And into our Best Practices doc...

proposals/security_and_privacy.md
mmccool: fixed

mmccool:
OK, I think all the reviews are addressed, ready to merge.

PR #186, opened 2019-03-27
WIP: Resynch TAG-review branch with master

mmccool:
WIP: Do not merge yet

Update master so it is in synch with changes in TAG-review branch (which are not in the core spec, just the supporting documents, but it is better to be consistent).


mmccool:
This is MOSTLY a good idea except for the change to the README, which we don't want to merge back into the master.   I will hold off on the merge until I figure out how to handle that.

mmccool:
Let me do this a different way...

Issue #178, opened 2019-03-25
Mark Example Sequences section as non-normative, fix diagrams


mmccool:
Currently the top of the section does not include the non-normative statement, but as the title includes the word "Examples" and most of the content is figures (which are considered non-normative...) it should be marked as such.

As an aside the terminology used in the headers in the diagrams are a bit mixed up.

Currently they are Application, Scripting API, WoT Server, and Device.
But this assumes the Device is a WoT Server.   The diagrams also 
say "Client" but "Server (Thing)".   They should be fixed to say "WoT Client" and "WoT Thing".

I personally think these headings should be...
       WoT Client                                             WoT Server
Application  - WoT Runtime  --- WoT Runtime - Application

The diagrams also use specific Scripting API calls which may be out of date.  Maybe we should be more generic and say "calls write property" rather than "calls thing.writeProperty()", etc.

There are also various "Simple WebSocket Binding" examples but we don't (unfortunately) have a defined subprotocol for that.   We can keep the examples but should make the status clear in the introduction to those sections (eg "one possible WebSocket binding").

mmccool:
Instead of "WoT Runtime" we might also want to use "Consumed Thing" and "Exposed Thing".

mmccool:
I am going to create a PR for this so we can discuss in the Arch call on March 28.

PR #189, opened 2019-03-27
Revise Section 11 (now Appendix A) Example Sequences to use consistent terminology

mmccool:
Revisions to section 11/appendix A (Example Sequences) to use consistent terminology

Addresses Issue https://github.com/w3c/wot-architecture/issues/178

Section 11 updates:
- class was set to "appendix informative" but it was not sticking because the id was also (incorrectly) given using "class=".   Fixed... this is now Appendix A, not Section 11
- Refer to correct operation names
- Say "a WebSocket" subprotocol, since we have not defined a standard one and there could be several
- Note that we are dealing with plain HTTP, not HTTPS (in the latter case there would be some additional protocol transactions to set up TLS, and even more if there was authentication, etc).
- Note that there is no explicit notification operation type, so the subprotocol for notification needs to be associated with the subscription operations.
- Update figures as described in issue (note: puml files modified, images regenerated using plantuml)

Other non-section 11 updates:
- Wrap the "Well-Known Operation Types" table in a figure so it can be referenced
- Add "unobserveproperty" to table to be consistent with the TD spec
- Expand tabs to 4 spaces in case there are any stragglers


mlagally:
Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fwot-architecture%2F&doc2=https%3A%2F%2Fcdn.staticaly.com%2Fgh%2Fmmccool%2Fwot-architecture%2Fsequences-updates%2Findex.html#property-interactions

Issue #192, opened 2019-03-28
Fix references


mmccool:
References to wot-security-best-practices and wot-security-testing-plan are broken, and references to other WoT documents are to the github repos and should point to the actual published location instead.

See PR https://github.com/w3c/wot-architecture/pull/193 for a fix to the TAG-review branch.
Some but not all of the same fixes should be applied to the master as well.

PR #193, opened 2019-03-28
Fix references to unpublished security drafts and TAG TD spec

mmccool:
It was pointed out to me that since the wot-security-best-practices and wot-security-testing-plan have not been published as official W3C documents the normal ReSpec auto-lookup mechanism would not work and the current W3C links explicitly included in the document were also broken (returning 404s).   It is also a irregular that we refer to the TD, BT, and SA specs using URLS pointing at github.

This patch:
- Refers to the editor's drafts (heads of the master branches) for wot-security-best-practices and wot-security testing-plan
- Uses respec lookup for the published version of wot-security as well as wot-thing-description, wot-binding-templates and wot-scripting-api
- Adds a reference (inside an ednote paragraph) to the TD review version of wot-thing-description
- Updates the links to github sites for official documents to use respec references instead.

Some of these edits, except for the reference to the TAG review version, should probably also be made to the master.

Addresses Issue https://github.com/w3c/wot-architecture/issues/192.
There are probably similar problems in the other documents (TD, etc) too...

### DIFF
--- a/index.html
+++ b/index.html
@@ -74,19 +74,34 @@
 				publisher : "IETF",
 				date : "03 July 2017"
 			},
-			"WOT-SECURITY-CONSIDERATIONS" : {
+/* ReSpec will pick up the citation it is officially published
+			"wot-security" : {
 				href : "https://www.w3.org/TR/wot-security/",
-				title : "Web of Things Security and Privacy Considerations",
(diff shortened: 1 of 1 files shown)


mlagally:
@mmccool Thanks for the fix!

Issue #195, opened 2019-04-02
Summary section 4.3 should must not mandate client server roles of specific blocks


mlagally:
Text and picture in section 4.3 constrains client and server roles on the device / cloud blocks. This is only informative, however it contradicts normative sections which don't impose client or server functionality on individual blocks (devices, servers, edge devices).

The figure and text needs to be updated to make this very clear.

k-toumura:
Currently, we use the word 'server' in following four meanings:
1. An entity which provide a (abstracted) Thing
    - In Terminology section: 'WoT server: An entity that exposes a network interface consistent with a WoT Thing Description. …'
2. A passive reactive component in communication
    - In 4.2.1: '… one device … has a server role that can accept a request from the other devices …'
    - Issue #146: 'use cases … don't require adopting a server model in the thing'
3. A computer, (virtual) machine, process. program, etc.
    - In 4.2.6: 'a model of a device … that resides on a cloud server …'
4. A entity which is in the Internet
    - @chachamimm mentioned in issue #188: 'Basically in Internet Server is the global network, and Client is the local network.'

In use case section, we use the term 'server' to refer to **1**, I think.
But we can't use the term 'WoT server' in use case section to clarify the meaning, because it is a WoT-specific terminology.
So I think it is helpful for define 'server'(and 'client') as **1** in the beginning of 4.2 Common Patterns.
Or, we can use 'Device/Application', 'Provider/User' or some other terms which do not imply the meaning of **2**,**3**, and **4** of above for describe the entity in Use Case section.

One of the drawback is we can't express difficulty of NAT and firewall traversal problem on Use Case section, because the problem is caused by a deployment pattern of active/passive component.(**2** of above listing).

PR #201, opened 2019-04-04
Update a figure in use case summary

k-toumura:
For issue #195.
I created an updated version of 'Figure 14 Use Case Overview' (not overwrite .png, but created new file).
@mlagally @mkovatsc Please review and comment if there are any problems.


k-toumura:
- Update [4.2 Common Patterns](https://w3c.github.io/wot-architecture/#sec-common-usecase-patterns) and [4.3 Summary](https://w3c.github.io/wot-architecture/#summary) to avoid enforcing server paradigm on devices.
  - In 4.2, use client/server role as roles in transport protocol layer, and show a device acts as both of roles.
  - Section 4.3 describes that application layer treat a device as provider of an abstract interface for interaction, regardless of roles in transport protocol layer.  The requirements in [Section 5.2.2 Devices](https://w3c.github.io/wot-architecture/#devices) can be derived from this.
- Minor editorial fix (closing tag, duplicated id)  
- update link of Figure 14.

PR #219, opened 2019-04-11
Move Section 6.2 to the end of 6

### LINKED ISSUE #188: Describe Server and Client at the section "6.2 WoT Architecture"
I have two opinions.

1) WoT Server and Wot Client  are described At the section "6.2 WoT Architecture". And Server and  Client are not described. But Server and Client is described at Fig 16 and Fig 17. I can't understand the reason why Server and Client is described. I think that these should be used the same terminology "WoT Server" and "WoT Client".

2) And I think that the relation of "Server" and "Client" is different in IoT and Internet. Basically in Internet Server is the global network, and Client is the local network. But in IoT Server is the local network, Client is the global network.  I think that WoT members are able to understand it, but other member may mot be able to understand. I think that "Server" and "Client" should be described in detail.

mkovatsc:
First step to fix #188 

Looking closer it turned out that we do not need a new Section 6.2, but the high-level view of Things, Consumers, and Intermediaries can be part of 6.1 Overview.

The "system configurations" are now 6.10.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1558,6 +1558,9 @@ a[href].internalDFN {
                 layer) information. More information about the TD can be
                 found in <a href="#sec-thing-description"></a>.
             </p>
+            <p class="ednote" title="TODO">
+                There is text incoming to explain the basics around Things, Consumers, and Intermediaries.
+            </p>
             <p>
                 <a href="#architecture-abstract"></a> shows an overview
                 of how these WoT concepts can be mapped to various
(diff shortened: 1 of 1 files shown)


mkovatsc:
Please merge, so that we can continue the work on 6.1 ~ 6.9 ( @mkovatsc ) and 6.10 ( @takuki and @mryuichi ).

mkovatsc:
https://cdn.staticaly.com/gh/w3c/wot-architecture/c64e9dfb5178b58ddbacd345c976b18e330bd2b0/index.html

PR #250, opened 2019-04-25
Update Figure 16 Linked Things

k-toumura:
- Add text 'Related' on gray arrow.
- Avoid overlapping green arrow on Thing A.

PR #244, opened 2019-04-25
Update Figures 27 and 28 (Servient implementation) and also 25

mmccool:
Resolves Issue https://github.com/w3c/wot-architecture/issues/232.   Note the figure numbering has changed though and I also made edits to a related figure (the building blocks figure) for consistency.

PR #242, opened 2019-04-23
Fix Web Thing text and figure and fix #230 (overview figure)

mkovatsc:
See #230 

As the overview reuses the Thing illustration from Section 6.3, I also fixed that section and figure.

PR #251, opened 2019-04-25
Figure revision and cleanup

mkovatsc:
Instead of writing a lot of text, I directly drew my change proposals for #244 and #250 .

I also aligned the colors of several figures, but not all yet; System Configuration and Deployment figures are still missing.

I removed the outdated PNGs.


mkovatsc:
[Preview](https://cdn.staticaly.com/gh/w3c/wot-architecture/mkovatsc-patch-figures-1/index.html?env=dev)

I did not touch the index.html, but now noticed we need to adjust some sizes/widths.

k-toumura:
There are multiple TDs in [Figure 18](https://github.com/w3c/wot-architecture/blob/2822711405c87cbf6f0e1b134b3301b05eef9d24/images/arch-overview.png). 
See PR #242, comment https://github.com/w3c/wot-architecture/pull/242#issuecomment-485997274, https://github.com/w3c/wot-architecture/pull/242#issuecomment-486542076.

mkovatsc:
From #242 

> 1. Multiple TDs
    I understand that you want to indicate that there may be more than one TD for things at the edge or in the cloud., however IMO in this overview picture it might be one level of detail too deep.
    Otherwise you could add a sentence just explaining, why there could be more than one.

More importantly they are also Intermediaries, Intermediaries for the connected Things. Thus multiple TDs at this level.

> 2. Color semantics
    Until now we used green icons for TDs, now they are orange. I'm not sure that we use these colors consistently all over the spec - please double-check.

Orange is also the preferred color by @sebastiankb in all his slides. Furthermore, there are no such figures in the TD spec.

mkovatsc:
I introduced another tweak: Behavior implementations as scripts and native laguage are shown outside WoT Runtime, as the runtime instantiates runtime objects (Exposed/Consumed Thing) inside, but their code is outside (also rather in design time).

mkovatsc:
Yet another update to reflect the existing text in Section 8 on system API.
Adjusted colors.

mlagally:
Approved to merge in call on 2.5.

PR #261, opened 2019-04-29
Align Section 8 (Fix #215)

### LINKED ISSUE #215: Align section 8 with improved section 6 from issue 188

mkovatsc:
Revised the text in Section 8 (Servient Implementation).

The text is already aligned with the figure updates in #251.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1986,6 +1986,8 @@ a[href].internalDFN {
                 A Protocol Binding is the mapping from an <a>Interaction Affordance</a> to concrete messages of a specific protocol such as HTTP [[!RFC7231]], CoAP [[!RFC7252]], or MQTT [[!MQTT]].
                 It informs the <a>Consumer</a> <em>how</em> to activate the <a>Interaction Affordance</a> through a network-facing interface.
                 The <a>Protocol Bindings</a> follow the Uniform Interface constraint of REST [[?REST]] to support interoperability.
+                Thus, not all communication protocols are eligible to implement <a>Protocol Bindings</a> for W3C WoT;
+                the requirements are given in the assertions below.
             </p>
             <p>
                 In the door example given in <a href="#sec-affordances"></a>, the <a>Protocol Binding</a> corresponds to the door handle at the level of knob vs lever,
@@ -2529,32 +2531,39 @@ a[href].internalDFN {
 
     <section id="sec-servient-implementation" class="informative">
         <h1>Servient Implementation</h1>
-        <p> A <a>Servient</a> is a software stack that implements the
-            WoT building blocks presented in the previous section. <a>Servients</a>
-            can host and expose <a>Things</a> and/or consume <a>Things</a>.
-            Depending on the <a>Protocol Binding</a>, <a>Servients</a> can perform in both server and client role,
+        <p>
+            A <a>Servient</a> is a software stack that implements the WoT building blocks presented in the previous section.
+            <a>Servients</a> can host and expose <a>Things</a> and/or consume <a>Things</a> (i.e., host <a>Consumers</a>).
+            Depending on the <a>Protocol Binding</a>,
+            <a>Servients</a> can perform in both server and client role,
             hence the portmanteau naming.
         </p>
-        <p> The previous section describes how the WoT
-            building blocks conceptually relate to each other and how they correspond to the
-            abstract <a href="#sec-wot-architecture">Web Architecture</a>.
-            When implementing these concepts, a
-            more detailed view is necessary that takes certain technical
-            aspects into account. The detailed architecture of a <a>Servient</a>
-            implementation using the (optional) <a>WoT Scripting API</a> building block is
-            shown in <a href="#architecture-implementation"></a>.
+        <p>
+            The previous section describes how the WoT building blocks conceptually relate to each other
+            and how they correspond to the abstract WoT Architecture (see <a href="#sec-wot-architecture"></a>).
+            When implementing these concepts,
+            a more detailed view is necessary that takes certain technical aspects into account.
+            This section presents the detailed architecture of a <a>Servient</a> implementation.
+        </p>
+        <p>
+            <a href="#architecture-implementation"></a> shows a <a>Servient</a> implementation using the (optional) <a>WoT Scripting API</a> building block.
             Here, the <a>WoT Runtime</a> is also a <a>Scripting Runtime</a> system that
             in addition to managing the WoT-specific aspects also interprets and executes the application scripts.
+            <a>Servients</a> with <a>WoT Scripting API</a> support usually run on powerful devices, edge nodes, or in the cloud.
+            There are, however, also <a>Scripting Runtimes</a> for resource-constrained systems.
+            Furthermore, the WoT Architecture does not limit the application-facing API of the <a>WoT Runtime</a> to JavaScript/ECMAScript [[ECMAScript]].
+        </p>
+        <p>
+            <a href="#native-impl"></a> presents an alternative <a>Servient</a> implementation without the <a>WoT Scripting API</a> building block.
+            The <a>WoT Runtime</a> may use any programming language for its application-facing API.
+            Usually, it is the native language of the <a>Servient</a> software stack,
+            for instance C/C++ for embedded <a>Servients</a> or Java for cloud-based <a>Servient</a>.
+            It may also be an alternative scripting langague such as Lua to combine the benefits of application scripts with low resource consumption.
         </p>
         <figure id="architecture-implementation">
             <img src="images/wot-thing-scripting.png" style="width: 80%;" />
             <figcaption>Implementation of a Servient Using the Optional WoT Scripting API</figcaption>
         </figure>
-        <p>
-            <a>Servients</a> with <a>WoT Scripting API</a> support usually run on powerful devices, edge nodes, or in the cloud.
-            There are, however, also <a>Scripting Runtimes</a> for resource-constrained systems.
-            Furthermore, the WoT Architecture does not limit <a>WoT Scripting API</a> definitions to JavaScript.
-        </p>
         <p>
             The role and functionality of each internal module shown
             in <a href="#architecture-implementation"></a> is given in
@@ -2572,15 +2581,21 @@ a[href].internalDFN {
                 which <a>Things</a>, <a>Consumers</a>, and <a>Intermediaries</a> are hosted on this component.
             </p>
             <p>
-                In the case the <a>Servient</a> utilizes the optional <a>WoT Scripting API</a> building block,
-                portable application scripts define the behavior.
+                <a href="#architecture-implementation"></a> focuses on <a>Servients</a> implementing the optional <a>WoT Scripting API</a> building block,
+                where portable application scripts written in JavaScript [[ECMAScript]] define the behavior.
                 They are executed in a <a>Scripting Runtime</a> system, which can be seen as either nested in or coinciding with the <a>WoT Runtime</a>.
-                They are portable, as they are written against the <a>WoT Scripting API</a> definitions,
-                and hence can be executed by any compatible <a>Servient</a>.
+                They are portable, as they are written against the common <a>WoT Scripting API</a> definitions,
+                and hence can be executed by any <a>Servient</a> featuring this building block.
                 This makes it possible to migrate certain application logic between system components,
                 for instance moving a <a>Consumer</a> from the cloud to an edge node to meet networking requirements,
                 or move an <a>Intermediary</a> to the cloud to fulfill growing resource demands.
                 The portability can also foster <a>Thing</a> customization by installing behavior following the app concept known from smartphones.
+            </p>
+            <p>
+                See <a href="#native-impl"></a> for behavior implementation without the <a>WoT Scripting API</a> building block.
+                In principle, any programming language and API can be used in order to define the behavior of a Thing,
+                as long as the <a>Interaction Affordances</a> are presented externally through a <a>WoT Interface</a>.
+                The adaption between application-facing API and the protocol stack is handled by the <a>WoT Runtime</a>.
             </p>
         </section>
         <section>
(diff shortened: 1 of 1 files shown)


mkovatsc:
[Rendered preview](https://cdn.staticaly.com/gh/w3c/wot-architecture/mkovatsc-patch-section-8/index.html?env=dev)

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
-                Using the Optional WoT Scripting API.
-            </figcaption>
+            <img src="images/wot-thing-scripting.png" style="width: 80%;" />
+            <figcaption>Implementation of a Servient Using the Optional WoT Scripting API</figcaption>
         </figure>
-        <p> The role and functionality of each additional module shown
+        <p>
+            <a>Servients</a> with <a>WoT Scripting API</a> support usually run on powerful devices, edge nodes, or in the cloud.
+            There are, however, also <a>Scripting Runtimes</a> for resource-constrained systems.
+            Furthermore, the WoT Architecture does not limit <a>WoT Scripting API</a> definitions to JavaScript.
```
k-toumura: This sentence is something contradicting with a sentence in [8.4 WoT Scripting API](https://cdn.staticaly.com/gh/w3c/wot-architecture/mkovatsc-patch-section-8/index.html?env=dev#wot-scripting-api) (l. 2631),
> The WoT Scripting API building block defines an ECMAScript API that ...

I think this sentence can be written as below: 
> Furthermore, the WoT Architecture does not limit _northbound WoT Runtime interface_ definition to ECMAScript.  

(implying existence of Native WoT Runtime and Native WoT API)

What do you think?

mlagally: Good catch. I'm not sure if the term "northbound" is known to all readers. Perhaps just call it "application-facing".

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
-            while WoT-specific Servient Implementation aspects are managed
-            by the WoT Runtime. 
-            <a>Consumed Thing</a> and <a>Exposed Thing</a>
-            objects provide abstract interfaces to the Things
-            described in related Thing Descriptions. Scripts can
-            operate on Things using these abstractions by performing
-            abstract Interactions while the details of the Protocol Bindings are
-            managed automatically by the WoT Runtime.
+            Here, the <a>WoT Runtime</a> is also a <a>Scripting Runtime</a> system that
+            in addition to managing the WoT-specific aspects also interprets and executes the application scripts.
```
mlagally: It would be worth clarifying that it is also possible to implement a WoT Servient in other (non-scripting) languages, such as C, C++, Java, C#, ...

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
-            <figcaption>Implementation of a Servient
-                Using the Optional WoT Scripting API.
-            </figcaption>
+            <img src="images/wot-thing-scripting.png" style="width: 80%;" />
+            <figcaption>Implementation of a Servient Using the Optional WoT Scripting API</figcaption>
         </figure>
-        <p> The role and functionality of each additional module shown
+        <p>
+            <a>Servients</a> with <a>WoT Scripting API</a> support usually run on powerful devices, edge nodes, or in the cloud.
+            There are, however, also <a>Scripting Runtimes</a> for resource-constrained systems.
```
mlagally: See comment above. We should not constrain it to "Scripting Runtimes". Other runtimes may even consume less resources, i.e. they can be deployed on smaller devices.

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
-                Servient actually does when a certain action is invoked)
-                and optionally other behaviors, such as control loops or
-                support for local user interfaces. 
-                The <a>WoT Scripting API</a> is one way to
-                implement an application using an application script,
-                and is described in more detail below. Application
-                scripts run inside a <a>Scripting Runtime</a> nested
-                inside the <a>WoT Runtime</a> but have access to objects and
-                interfaces defined by <a>WoT Runtime</a>.
+            <h2>Behavior Implementation</h2>
```
mlagally: Suggest to rename the title to "Implementation of the behavior of a thing"

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
+            <p>
+                The behavior defines the overall application logic and includes
+                autonomous behavior of <a>Things</a> (e.g., sampling of sensors or control loops for actuators),
+                the handlers for <a>Interaction Affordances</a> (i.e., the concrete actions taken when an affordance is activated),
+                <a>Consumer</a> behavior (e.g., controlling a <a>Thing</a> or realizing mashups), and
+                <a>Intermediary</a> behavior (e.g., simply proxying a <a>Thing</a> or composing virtual entities).
+                The behavior implementation within a <a>Servient</a> defines
```
mlagally: reword to: implementation of the "behavior" ?

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
+                which <a>Things</a>, <a>Consumers</a>, and <a>Intermediaries</a> are hosted on this component.
+            </p>
+            <p>
+                In the case the <a>Servient</a> utilizes the optional <a>WoT Scripting API</a> building block,
+                portable application scripts define the behavior.
```
mlagally: Is the "portability" aspect relevant here?

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
+                The behavior defines the overall application logic and includes
+                autonomous behavior of <a>Things</a> (e.g., sampling of sensors or control loops for actuators),
+                the handlers for <a>Interaction Affordances</a> (i.e., the concrete actions taken when an affordance is activated),
+                <a>Consumer</a> behavior (e.g., controlling a <a>Thing</a> or realizing mashups), and
+                <a>Intermediary</a> behavior (e.g., simply proxying a <a>Thing</a> or composing virtual entities).
+                The behavior implementation within a <a>Servient</a> defines
+                which <a>Things</a>, <a>Consumers</a>, and <a>Intermediaries</a> are hosted on this component.
+            </p>
+            <p>
+                In the case the <a>Servient</a> utilizes the optional <a>WoT Scripting API</a> building block,
```
mlagally: If a Servient implements the optional ...

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
+                portable application scripts define the behavior.
+                They are executed in a <a>Scripting Runtime</a> system, which can be seen as either nested in or coinciding with the <a>WoT Runtime</a>.
```
mlagally: I think the important aspect is portability and not the implementation language, i.e. I'm not clear why we distinguish between Scripting runtime and WoT runtime.

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
+                and use internal interfaces to realize the lifecycle management.
+                The <a>WoT Scripting API</a> includes one possible implementation of the LCM functionality.
+            </p>
+            <p>
+                Southbound, the <a>WoT Runtime</a> interfaces with the protocol stack implementation of the <a>Servient</a>
+                and decouples the behavior implementation from the details of the <a>Protocol Bindings</a>.
+                The <a>WoT Runtime</a> usually also interfaces with the underlying system,
+                for instance, to access local hardware or even proprietary communication protocols.
+                Both interfaces are implementation-specific, but must follow the <a>Thing</a> abstraction,
+                so that the behavior implementation need not depent on system-specific API calls.
```
mlagally: I always stumble over the term "behavior implementation".

Please consider rewording to: 
- "behavior" implementation 
or
- implementation of the behavior

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
-                external disclosure of this data. Private security data
-                is used transparently by the <a>Protocol Binding</a> to
-                authorize and protect the integrity and confidentiality
-                of interactions.</p>
+            <h2>Scripting Runtime</h2>
+            <p>
+                When the <a>WoT Scripting API</a> is used, the <a>Servient</a> requires a language runtime to actually manage and execute scripts.
+                This <a>Scripting Runtime</a> is part of the <a>WoT Runtime</a>, but is functionally distinct from it.
+                While the <a>WoT Runtime</a> implements the <a>Thing</a> abstraction, provides lifecycle management, and interfaces with the protocol stack and system,
+                the <a>Scripting Runtime</a> interprets the scripting language executes the application scripts (e.g., ECMAScript [[?ECMASCRIPT]]).
```
mlagally: See comment above on byte-code interpreters. The architecture should not preclude portable *applications*, that may be written in other languages.

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
+                and providing their autonomous behavior.
+            </p>
+            <p>
+                The <a>Consumed Thing</a> abstraction represents a remotely hosted <a>Thing</a> for <a>Consumers</a> that needs to be accessed using a communication protocol.
+                <a>Consumed Things</a> are proxy objects or stubs.
+                The behavior implementation is restricted to reading their metadata and activating their <a>Interaction Affordances</a> as described in the corresponding <a>TD</a>.
+                <a>Consumed Things</a> can also represent system features such as local hardware or devices behind proprietary or legacy communication protocols.
+                In this case, the <a>WoT Runtime</a> must provide the necessary adaptation between system API and <a>Consumed Thing</a>.
+                Furthermore, it must provide corresponding <a>TDs</a> and make them available to the behavior implementation,
+                for instance, by extending a discovery mechanism.
```
mlagally: What does this sentence mean? How do you extend a discovery mechanism?

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
+            </p>
+            <p>
+                When using the <a>WoT Scripting API</a>, <a>Exposed Thing</a> and <a>Consumed Thing</a> are JavaScript objects,
+                which can be created, operated on, and descroyed by application scripts.
```
mlagally: destroyed

index.html
mkovatsc: I will rephrase, so that "WoT Scripting API" refers to the defined ECMAScript API and "other scripting API" are allowed by the architecture.

Also, I will replace north-/southbound.

index.html
```diff
@@ -2532,183 +2532,169 @@ <h1>Servient Implementation</h1>
...
-                 <a>Consumed Things</a> is realized in the same manner as that
-                 of <a>Exposed Things</a>.
-             </p>
+            <h2>Private Security Configuration</h2>
+            <p>
+                Private security metadata is also conceptually
+                managed by the <a>WoT Runtime</a> but is intentionally not made
+                directly accessible to the application. In fact, in the
+                most secure hardware implementations, such private
+                security data is stored in a separate, isolated memory
```
mlagally: memory -> entity ?

index.html
mkovatsc: That is the point of 8.10.1 and why we explicitly state this section explains the variant with the optional Scripting API.

index.html
mkovatsc: I will include a reference to the non-scripting alternative in the Section 8 introduction text.

index.html
mkovatsc: It is intended to reflect the label of the box in the figure.

index.html
mkovatsc: It is not limited to Things, it is behavior implementation in general. It once was "application", but was changed to "behavior" and hence "behavior implementation" in this context. I was never so happy about this wording.

index.html
```diff
@@ -2723,66 +2709,63 @@ <h2>Protocol Bindings</h2>
...
-            <p> A device may also be physically external to a <a>Servient</a>,
-                but connected via a proprietary protocols or a protocol
-                not eligible for direct support as part of the WoT. In
-                this case, the implemented runtime environment may
+                behavior implementations through the <a>Thing</a>
+                abstraction, as if they were accessible over a communication
+                protocol. In this case, the <a>WoT Runtime</a> should
+                enable the behavior implementation to instantiate <a>Consumed Things</a>
+                that internally interface with the system instead of the protocol stack.
+                This can be done by listing such system Things in the discovery results.
```
mlagally: What is the implication of being a "system Thing"? IMO it means that there does not exist a TD but only an object in the runtime. 
"Discovery" probably means local introspection.

index.html
mkovatsc: Having multiple standard APIs in different languages, even if it is just two, already contradicts the portability argument.

index.html
```diff
@@ -2723,66 +2709,63 @@ <h2>Protocol Bindings</h2>
...
+            </p>
+            <p>
+                A device may also be physically external to a <a>Servient</a>,
+                but connected via a proprietary protocol or a protocol
+                not eligible as <a>WoT Interface</a>. In
```
mlagally: This may be a question by some readers: What minimum constraints does a protocol have to satisfy to be "eligible"?
Is using a URI scheme sufficient?

index.html
mkovatsc: This needs to be commented by @mmccool and @zolkis, who established this distinction and wrote the original section and text on this.

index.html
```diff
@@ -2723,66 +2709,63 @@ <h2>Protocol Bindings</h2>
...
-                an application; and those that do not use a <a>WoT Runtime</a>
-                but are implemented using a separate software stack. The
-                second case can still be incorporated into a WoT system
-                if they use a WoT-supported protocol and a Thing
-                Description can be provided for such devices or services.
+                northbound APIs for the <a>WoT Runtime</a> or provide the
+                behavior implementation in a native programming language.
+            </p>
+            <p>
+                Furthermore, devices or services unaware of W3C WoT can still be consumed,
```
mlagally: unaware of WoT = not implementing any building block? Does this have any implications to a consumer?

index.html
mkovatsc: You mean putting the word behavior in quotes?

I don't understand why this issue only arises now and not when the term was introduced and caused big ripples throughout the document, nor when I tried to discuss if "behavior" is the term we want use.

mlagally:
@mkovatsc This looks very good.
I have added several questions and comments, that are intended to help sorting out ambiguities.

index.html
mkovatsc: Where is the relation to portability here?

Machine code with dynamic linking would be just as "portable" assuming everyone has that execution environment. The portability mentioned in other places is mainly about the common API (cf. Web browser APIs) and a single execution environment (cf. Web apps in JavaScript).

Will fix the missing "and" and remove the ECMAScript reference.

mkovatsc:
Thanks for the reviews, will push an update soon.

index.html
mkovatsc: Rephrased.

index.html
mkovatsc: Added reference to the WoT Scripting API discover() method as illustration.

index.html
mkovatsc: Fixed.

index.html
mkovatsc: I would leave it with memory, as it referres to a concrete implementation strategy (e.g., with secure hardware elements, which offer such memory).

index.html
mkovatsc: Rephrased: in the results of the discovery mechanism provided by the application-facing <a>WoT Runtime</a> API

index.html
mkovatsc: There are multiple subsections on this, which I now linked.

index.html
mkovatsc: No, as long as the second part of the sentence is fulfilled.

mkovatsc:
I tried to address all the comments. Please review again (also see replies to the comments).

mlagally:
approved in call on April 2nd

Issue #217, opened 2019-04-11
Align section 10 with improved section 6 from issue 188


mlagally:

mmccool:
See PR https://github.com/w3c/wot-architecture/pull/271

mkovatsc:
Was merged via CR-candidate-1.

PR #271, opened 2019-05-01
Update Section 10 (Issue #217)

mmccool:
Addresses Issue https://github.com/w3c/wot-architecture/issues/217.
Actually, there was nothing really to fix, the existing text was consistent with the new definitions.   In particular WoT Server and WoT Client were never used in the original text.   However, I did make a few small fixes and replaced some instances of "users" with "Consumers".


mlagally:
approved in call on May 2nd

Issue #209, opened 2019-04-11
Section 8.10.2 Existing Device Described with a Thing Description is too restrictive


mlagally:
"8.10.2 Existing Device Described with a Thing Description" contains the following sentence:

It is also possible to support existing IoT devices or services in the WoT and integrate them with other WoT Things by writing a Thing Description for them. __This can only be done however if the target device is using protocols eligible for WoT Protocol Bindings.__

This is too restrictive and limits the usefulness of WoT. It is possible to define protocol bindings for new protocols, the system is extensible. It is also possible to use a device from WoT by creating a protocol adapter on a gateway / proxy.

mmccool:
See PR https://github.com/w3c/wot-architecture/pull/273

mkovatsc:
Was merged via CR-candidate-1.

PR #273, opened 2019-05-01
Clarify use of existing devices (Issue #209)

mmccool:
Addresses Issue https://github.com/w3c/wot-architecture/issues/209.


mlagally:
@mmccool In PR https://github.com/w3c/wot-architecture/pull/271 you did an update to the figures as well. Are they different?

mlagally:
approved in call on May 2nd

PR #287, opened 2019-05-03
Mark up all assertions

mmccool:
Mark up all RFC2119 assertions so they can be recovered to generate an implementation report.
A few assertions have been reworded to avoid compound sentences or ambiguous pronoun usage.


mlagally:
I checked the rendered version at:
https://cdn.staticaly.com/gh/mmccool/wot-architecture/assertion-markup/index.html?env=dev
Merging since the changes are purely editorial

PR #288, opened 2019-05-05
Remove direct refs to sec best practices and testing docs

mmccool:
Addresses issue https://github.com/w3c/wot-architecture/issues/284.   Also a few other editorial changes to avoid using citations as nouns, which is considered bad form (just to the citations to wot-security, there may be others that do this, but we can defer fixing these until after CR, it is an purely editorial issue).

### DIFF
--- a/index.html
+++ b/index.html
@@ -71,7 +71,7 @@
                 publisher : "IETF",
                 date : "03 July 2017"
             },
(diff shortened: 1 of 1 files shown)


mkovatsc:
Looks good to merge.

mkovatsc:
Please merge so we can continue working without tons of merge conflicts.

Issue #279, opened 2019-05-02
update section: 8.9.2 Existing Device Described with a Thing Description


mlagally:
Remove:
Providing a Thing Description for a device also only allows it to be used in a server role, with its Thing Description being consumed by other Things. Using such a device as a Consumer will only in general be possible if the device can be modified to consume Thing Descriptions.

mmccool:
Resolved in PR https://github.com/w3c/wot-architecture/pull/289.

Issue #276, opened 2019-05-02
Add text to section "Existing Device Described with a Thing Description"


mlagally:
Indicate manual creation of TD.

Add a sentence describing the mechanism to create a TD.

mmccool:
See PR https://github.com/w3c/wot-architecture/pull/289.

PR #289, opened 2019-05-05
Update discussion of support for existing devices (issues 279 and 276)

mmccool:
- Simplify paragraph describing devices that cannot be modified, avoiding implying that they must be servers (right now, TDs can only describe servers, but in the future this might not be the case).  Addresses issue https://github.com/w3c/wot-architecture/issues/279
- Discuss means by which TDs may be created, to be consistent with the figure.  Addresses issue https://github.com/w3c/wot-architecture/issues/276


mkovatsc:
Looks good to merge, except for `<em>WoT <a>Things</a></em>`. I think it is not good to somewhat locally redefine an already defined term... Just say `<a>Things</a>`, which already points to our definition.

mkovatsc:
Please merge. For some reason changes to the lines changed here already caused a conflict for #291.
After merging, I can address the WoT Thing issue by yet again resolving the conflict...

Issue #278, opened 2019-05-02
Remove "Scripting Runtime" as a term, use execution environment


mlagally:
Delete it from terminology section.
Remove section 8.3
Call out the sandboxing aspect of the application execution environment in section 8.2
update intro text in section 8

zolkis:
There is a distinction between WoT Runtime (that implements the WoT network interface) and a Scripting Runtime (which is an optional part of the WoT Runtime and provides runtime functionality to run scripts).
Sandboxing can be part of Scripting Runtime, but usually one Scripting Runtime is one sandbox.
Removing the term Scripting Runtime will result in the earlier situation of confusing the two.

mkovatsc:
The main point of the discussion was that, even when no scripting is used and there is no scripting runtime, the WoT Runtime must still be able to instantiate the ExposedThing and ConsumedThing software instances and provide an API toward the application code. In this sense, there is no point to emphasize a scripting runtime system that may or may not be used as part of the WoT Runtime.

zolkis:
> and provide an API toward the application code

No, this was not part of the discussions in the WG. Is that now suddenly a mandatory requirement (you used "must")? If yes, what API must that be?

I don't think there is any mandatory requirement that a WoT Runtime MUST expose any kind of API towards applications. Would you give a source to where was this even discussed as a requirement?
Even the standardized Scripting API is optional, now you suddenly want to make exposing a non-standard application interface a mandatory part of implementations???

What the discussions have been about in my interpretation: 
- a WoT Runtime is the software stack needed to instantiate Things that can communicate via the WoT network interface using WoT interactions. That SW stack can be an OS + libraries + applications flashed to the HW or otherwise provisioned: but we never specified how should that happen except that the Scripting API is one way to implement applications (that is standardized in the WG, according to the charter).
- the WoT Runtime includes handling security metadata (in scope for the WG work) and provisioning mechanism (out of scope for the WG work).

Since there was a confusion during the discussions between WoT Runtime and scripting runtimes (such as an ECMAScript runtime, or Lua runtime or Python runtime), it is even more likely that outsider readers might get confused, too, so we felt the need to give definitions for both in the Architecture document. 

>  there is no point to emphasize a scripting runtime system that may or may not be used as part of the WoT Runtime

That I agree with: this relationship can be indeed omitted, but the roles should be defined clearly and separately.
[...]

PR #291, opened 2019-05-05
Fix #277 and #278

### LINKED ISSUE #277: Remove other scripting language environments from the WoT scripting API section
Make clear that this is only ECMA-script:

> The WoT Scripting API is an optional "convenience" building block in WoT that eases IoT application development and is implemented using a Scripting Runtime (such as a Python, Lua or ECMAScript implementation) that is used in combination with a WoT Runtime. The WoT Scripting API enables writing WoT applications by user-defined scripts.

mkovatsc:
* Made clear that WoT Scripting API is only ECMA-script
* Deleted Scripting Runtime from terminology section.
* Removed Section 8.3
* Called out the sandboxing aspect of the application execution environment in section 8.2
* Updated intro text in section 8


index.html
```diff
@@ -424,8 +424,8 @@ <h1>Terminology</h1>
             <dt>
                 <dfn>Execution Environment</dfn>
             </dt>
-            <dd>A sandbox within the WoT Runtime that isolates scripts
-                running on the same Servient.</dd>
+            <dd>A sandbox within the WoT Runtime that isolates application logic
```
zolkis: I would avoid saying "sandbox" and I would define the term "Execution Context" rather than "Execution Environment". Only one agent is running in one context, there could be multiple contexts but one "running execution context". Usually an execution context has associated an execution environment (global or local), eventual global objects, intrinsics, call stack, callback queue etc so it's a larger concept.

mkovatsc: Execution Context is too scripting-specific. I think for the general meaning we need here, Execution Environment is fine (compare, for instance, with Trusted Execution Environment)

zolkis: Execution Context and also Execution Environment are _supposed to be_ code specific, of course.
They are not scripting-specific, as you have them with native code as well.

Environment is a small part of an execution context.

index.html
```diff
@@ -2512,22 +2507,20 @@ <h3>WoT Scripting API</h3>
...
                 edge node.
             </p>
-            <p> The non-normative <a>WoT Scripting API</a> document defines
-                the structure and algorithms of the programming
-                interface that allows scripts to discover, process
-                (retrieve) and produce (define) <a>Thing Descriptions</a>. The
-                implementation of the <a>WoT Scripting API</a> instantiates local
+            <p> The non-normative <a>WoT Scripting API</a> document [[?wot-scripting-api]
+                defines the structure and algorithms of the programming
+                interface that allows scripts to discover, fetch and process,
```
zolkis: There is no fetch() in Scripting API any more.

index.html
```diff
@@ -2512,22 +2507,20 @@ <h3>WoT Scripting API</h3>
...
+                and produce (i.e., define and publish) <a>WoT Thing Descriptions</a>. The
```
zolkis: there is no `publish()` but `expose()`

index.html
```diff
@@ -2512,22 +2507,20 @@ <h3>WoT Scripting API</h3>
...
+                runtime system of the <a>WoT Scripting API</a> instantiates local
                 objects that act as an interface to other <a>Things</a> and
-                represents their <a>Interaction Affordances</a> (<a>Properties</a>,
+                their <a>Interaction Affordances</a> (<a>Properties</a>,
                 <a>Actions</a>, and <a>Events</a>). It also allows scripts to expose
-                <a>Things</a>, i.e. to specify and implement <a>Interaction
+                <a>Things</a>, that is, to define and implement <a>Interaction
                 Affordances</a> and publish a <a>Thing Description</a>.
             </p>
             <p> The standardized <a>WoT Scripting API</a> provides a
```
zolkis: this paragraph could be removed, as it does not add much here

index.html
```diff
@@ -2624,12 +2618,17 @@ <h2>WoT Runtime</h2>
             <p>
                 Technically, the <a>Thing</a> abstraction and its <a>Interaction Model</a> is implemented in a runtime system.
                 This <a>WoT Runtime</a> maintains the execution environment for the behavior implementation and is able to expose and/or consume <a>Things</a>,
```
zolkis: execution context

index.html
```diff
@@ -2624,12 +2618,17 @@ <h2>WoT Runtime</h2>
...
-                The <a>WoT Runtime</a> must provide an API for the behavior implementation that follows the <a>Thing</a> abstraction.
-                The optional <a>WoT Scripting API</a> building block shown in <a href="#architecture-implementation"></a> defines such an application-facing interface.
+                The <a>WoT Runtime</a> must provide an API for the behavior implementation.
+                The optional <a>WoT Scripting API</a> building block shown in <a href="#architecture-implementation"></a>
+                defines such an application-facing interface that follows the <a>Thing</a> abstraction.
                 See <a href="#native-impl"></a> for alternative APIs.
+                In general, application logic should be executed in isolated <a>Execution Environments</a>
+                that prevent unauthorized access to the management aspects of the <a>WoT Runtime</a>,
+                in particular the private security configurations.
+                In multi-tenant <a>Servients</a>, additional <a>Execution Environment</a> isolation is required for the different tenants.
```
zolkis: an execution environment might be shared, it is the execution contexts that are isolated

index.html
```diff
@@ -2624,12 +2618,17 @@ <h2>WoT Runtime</h2>
             <p>
                 Technically, the <a>Thing</a> abstraction and its <a>Interaction Model</a> is implemented in a runtime system.
                 This <a>WoT Runtime</a> maintains the execution environment for the behavior implementation and is able to expose and/or consume <a>Things</a>,
-                and hence must be able to process, serialize, serve, and fetch <a>WoT Thing Descriptions</a>.
+                and hence must be able to fetch, process, serialize, and serve <a>WoT Thing Descriptions</a>.
             </p>
             <p>
-                The <a>WoT Runtime</a> must provide an API for the behavior implementation that follows the <a>Thing</a> abstraction.
-                The optional <a>WoT Scripting API</a> building block shown in <a href="#architecture-implementation"></a> defines such an application-facing interface.
+                The <a>WoT Runtime</a> must provide an API for the behavior implementation.
```
zolkis: Is this normative? If yes, you must define what kind of API is normative there. 
But this should not be normative, since a WoT Runtime may contain its applications (Things, behaviors) and manufacturers should be free to do so, without the requirement to expose non-specified APIs.

index.html
```diff
@@ -2624,12 +2618,17 @@ <h2>WoT Runtime</h2>
...
+                The optional <a>WoT Scripting API</a> building block shown in <a href="#architecture-implementation"></a>
+                defines such an application-facing interface that follows the <a>Thing</a> abstraction.
                 See <a href="#native-impl"></a> for alternative APIs.
+                In general, application logic should be executed in isolated <a>Execution Environments</a>
```
zolkis: An Execution Environment is not necessarily isolated. Then, you should specify what level of isolation do you mean.

index.html
mkovatsc: it is about the functionality in general. Would "retrieve" make you happier?

mkovatsc: There must be a functionality to reference a TD by URI, no?

zolkis: OK, since fetch() is there in the scripting runtime actually, in the general sense we can leave fetch there if it's not explicitly said it's a requirement for the API itself. From the current text I deduce this is a mandatory requirement for the Scripting API, which should not be - but it's a mandatory requirement to the Scripting Runtime (one of its global objects exposed to the Scripting API).

index.html
zolkis: Maybe you can leave `publish()` and maybe we need to change `expose()` to `publish()` in the Scripting API.

zolkis:
A generic comment: if the Scripting API is ECMAScript indeed, we SHOULD use the runtime terms (execution context, environment, etc) from ECMAScript.

index.html
mkovatsc: Execution context is mainly used in the context of JavaScript. There it even means something different, as it includes the current evaluation state of code. Will not change.

index.html
mkovatsc: Won't fix (see above)

index.html
mkovatsc: No RFC 2119 langauge here. Still, there is naturally always an API, otherwise no application logic can be specified. This can and often is an API at compile time when no scripting is involved.

index.html
mkovatsc: When the WoT Scripting API relies on a fetch() funtion in the runtime, it should make it part of the standardized API. Thus, it would again be part of the WoT Scripting API, even when not implemented by WoT-specific objects.

index.html
zolkis: In this PR you "make clear WoT API is only ECMAScript". Then please use the runtime terms defined in the [ECMAScript spec](http://www.ecma-international.org/ecma-262/9.0/index.html#sec-executable-code-and-execution-contexts) and don't assume everyone understands the same thing as you on "Execution Environment".

index.html
zolkis: Should fix (see above).

index.html
mkovatsc: I will change to "expose". This term fits in the API as it is about exposing the Thing and not only about making its TD available (~publish).

index.html
mkovatsc: Removing.

index.html
zolkis: Then you must specify what kind of API do you mean, at least on functional requirements terms. Saying "there is naturally always an API" doesn't mean anything, since between every code module there is an API, but that is no good reason to talk about an API in a normative document without specifics.

index.html
zolkis: No, the Scripting TF disagrees with you on this one (will provide link to the calls where this has been discussed). Fact is that fetch() is not any more part of the Scripting API.

mkovatsc:
Reasonable comments were addressed.

mlagally:
@zolkis, @mkovatsc Please participate in tomorrow's Architecture call (8am-10am CEST) to discuss and resolve any remaining issues.

mlagally:
@mkovatsc Unfortunately there's currently a merge conflict - can you please resolve?

mkovatsc:
Resolved.

mkovatsc:
Resolved again.

mlagally:
approved to merge in call on 7.5.

Issue #260, opened 2019-04-29
Terminology definition of Application


mkovatsc:
After several iterations on the text, the WoT-specific definition of "Application" does not really apply anymore. It is somewhat in conflict with "Consumer" and "Behavior".

I checked the text and would recommend to simply delete its definition from the Terminology section. The common understanding of "application" fits to the text and the meaning we want.

mlagally:
agreed in call on May 2nd.

PR #292, opened 2019-05-05
Remove Application definition from Terminology

mkovatsc:
See #260, fixed.


zolkis:
Now we don't have WoT Applications any more? So what do we have?

mlagally:
As agreed in last week's call. Merging.

Issue #182, opened 2019-03-25
Add a definition of Proxy


mmccool:
Since we have a definition of Digital Twin.  Definition should be consistent with discussion in 9.1.2.

mlagally:
call on 4.4.:
Matsukura-san will provide a PR.

mkovatsc:
On the conceptional level (Thing, Consumers), we now have Intermediaries.
See #221

mkovatsc:
Please clarify if we still need "proxy".

mlagally:
call on 25.4.:
Defer - let's incorporate the terminology updates from section 6 into the rest of the document first.

mkovatsc:
See #293.
We do not need such a definition. I propose closing this issue.

PR #293, opened 2019-05-05
Revise usage of term proxy

mkovatsc:
Revisting #182, I made a pass and clarified all occurrences of the term "proxy". As a result, we definitely do not need a definition of the term.

I also fixed a statement in 9.2.4 to clarify that Consumers discover Things through the directory and then connect to the Things.


mlagally:
Looks good.

PR #300, opened 2019-05-06
Editorial chapter 6

mlagally:


index.html
```diff
@@ -1627,7 +1638,7 @@ <h2>Affordances</h2>
             <p>
                 A central aspect in W3C WoT is the provision of machine-understandable metadata (i.e., <a>WoT Thing Descriptions</a>).
                 Ideally, such metadata is self-descriptive, so that <a>Consumers</a> are able to identify
-                <em>what</em> capabilities a <a>Thing</a> provides and <em>how</em> to use the provided capabilities.
+                <em>which</em> capabilities a <a>Thing</a> provides and <em>how</em> to use the provided capabilities.
```
mkovatsc: Please leave "what". "Which" has the notion of a well-known list from which must be picked.

index.html
```diff
@@ -1809,18 +1824,19 @@ <h2>Hypermedia Controls</h2>
...
-                A hypermedia control is the machine-understandable serialization of <em>how</em> to activate an affordance.
+                But also machines can follow links in a meaningful way, when the Web link is further described 
+                by a relation type and a set of target attributes.
+                A hypermedia control is the machine-understandable description of <em>how</em> to use an affordance.
                 Hypermedia controls usually originate from a Web server and are discovered in-band while a Web client is interacting with the server.
                 This way, Web servers can drive clients through Web applications dynamically,
                 by taking their current state and other factors such as authorization into account.
                 This is opposed to out-of-band interface descriptions that need to be preinstalled or hardcoded into clients
-                (e.g., RPC, WS-* Web services, HTTP services with fixed URI-method-response definitions).
+                (e.g., RPC, WS-* Web services, HTTP+REST services with fixed URI-method-response definitions).
```
mkovatsc: Please remove REST. There is no REST when the URIs are hard-coded and resources are typed through out of band information.

I know that too many people casually  call anything HTTP based REST(ful), but it is wrong.

index.html
```diff
@@ -1922,30 +1938,29 @@ <h3>Forms</h3>
...
+                            Nevertheless, all-lowercase URIs SHOULD be used for extension operation types.
                         </span>
                     </li>
                 </ul>
                 <p>
                     <span class="rfc2119-assertion" id="arch-op-request-method">
-                    The request method MUST identify one method of
-                    the standard set of the protocol identified by the
+                    The request method MUST identify one operation type of
+                    the standard set of operation types of the protocol identified by the
```
mkovatsc: The operation type (op) is different from the method.
Please revert to "method".

index.html
```diff
@@ -2061,7 +2076,7 @@ <h3>Forms</h3>
             <h2>Protocol Bindings</h2>
             <p>
                 A Protocol Binding is the mapping from an <a>Interaction Affordance</a> to concrete messages of a specific protocol such as HTTP [[!RFC7231]], CoAP [[!RFC7252]], or MQTT [[!MQTT]].
-                It informs the <a>Consumer</a> <em>how</em> to activate the <a>Interaction Affordance</a> through a network-facing interface.
+                It informs the <a>Consumer</a> <em>how</em> to use the <a>Interaction Affordance</a> through a network-facing interface.
```
mkovatsc: Please revert to activate.
We checked this wording during the hypermedia workshop with IETF people.

index.html
```diff
@@ -2081,11 +2096,13 @@ <h3>Hypermedia-driven</h3>
                         href="#sec-hypermedia-controls"></a>) to be self-descriptive on how to activate the Interaction Affordance.
                     </span>
                     <span class="rfc2119-assertion" id="arch-hypermedia-origin">
-                        The hypermedia controls MUST originate from the authority managing the Thing providing the corresponding Interaction Affordance.
+                        The hypermedia controls MUST originate from the authority managing the Thing which is providing the corresponding Interaction Affordance.
```
mkovatsc: Please use "that" instead of "which", as it is a restrictive subclause

index.html
```diff
@@ -2122,17 +2139,17 @@ <h3>Media Types</h3>
...
                     for JSON [[!RFC8259]] or
                     <code>application/cbor</code>
-                    for CBOR [[!RFC7049]].
+                    for CBOR [[!RFC7049]]. They are managed by IANA. 
                 </p>
                 <p>
                     Some Media Types might need additional parameters to fully specify the representation format used.
                     Examples are <code>text/plain; charset=utf-8</code> or <code>application/ld+json; profile="http://www.w3.org/ns/json-ld#compacted"</code>.
-                    This needs to be taken into account in particular when describing data to be sent to <a>Things</a>.
+                    This needs to be taken into account in particular when describing data to be sent to, or received by <a>Things</a>.
```
mkovatsc: Please remove "received by", because in that case the Thing must apply and set the content type in the message. A media Type hint would be enough.
When sending to Things, the Consumer must know the exact content type to formulate a correct request. A media type hint would not be enough. Hence "in particular"  for sending to Things.

index.html
```diff
@@ -2144,7 +2161,7 @@ <h3>Media Types</h3>
                     semantics for its elements (e.g., XML, JSON, CBOR). 
                     <span class="rfc2119-assertion" id="arch-schema">Thus,
                         the corresponding Interaction Affordances SHOULD
-                        declare a data schema to provide more detailed
+                        declare a <em>data schema</em> to provide more detailed
                         semantic metadata for the data exchanged.</span>
```
mkovatsc: Could you fix this to "syntactic metadata", please?

index.html
```diff
@@ -2174,29 +2196,29 @@ <h2>WoT System Components and their Interconnectivity</h2>
 
             <p>On the other hand, <a>Consumers</a> are always implemented by 
                <a>Servients</a>, as they must be able to process the <a>Thing Description</a> (TD) 
-               format and must have a protocol stack that can be configured through 
+               format and must have a protocol stack that can be configured to use a thing through
```
mkovatsc: The protocol stack itself needs to be configured. The configuration information is given in the Protocol Binding. Hence, "configured through" or "by" Protocol Binding information.

mlagally: I agree that the protocol stack needs to be configured, but that's still expressed here.
My change explains the purpose of configuring the protocol stack, i.e. "to use a thing".
IMHO the change adds clarity.

index.html
mlagally: good catch

index.html
mlagally: hmmm. I think the term "activate" is not ideal, since it implies that there is some activity happening at the thing, which is not the case if you read a property.
Using can mean to activate, to read, to subscribe etc, ...

mkovatsc: "Using" could also just mean to learn what choices exist.

index.html
mlagally: ok

index.html
mlagally: ok

index.html
mkovatsc: "configured to use a *T*hing" sounds to me like just giving a target... maybe "configured for a Thing through"?

index.html
mlagally: ok

index.html
mlagally: 👍

index.html
mlagally: "interact with" ?

mkovatsc: "activate" -- as I said we already invested time into this.
This terminology you can find for instance in texts by Mike Amundsen (the RESTful Web APIs O'Reilly book and more...).

mlagally:
Requested changes were addressed and the CR-candidate-2 was updated.

mlagally:
approved to merge in call on 7.5.

PR #302, opened 2019-05-06
Editorial chapter 8

mlagally:


index.html
```diff
@@ -2608,8 +2608,11 @@ <h3>WoT Security and Privacy Guidelines</h3>
     <section id="sec-servient-implementation" class="informative">
         <h1>Servient Implementation</h1>
         <p>
-            A <a>Servient</a> is a software stack that implements the WoT building blocks presented in the previous section.
-            <a>Servients</a> can host and expose <a>Things</a> and/or consume <a>Things</a> (i.e., host <a>Consumers</a>).
+            As described in section <a href="sec-WoT-servient-architecture-high-leve"></a>
```
k-toumura: This might be: `<a href="#sec-WoT-servient-architecture-high-level"></a>`

mkovatsc: Should also be all lower case (cf. WoT)

mkovatsc:
Not in Section 8, but this is an open PR: There are occurances of "WoT Building Blocks", where building blocks is capitalized. Since it is not a defined term under Terminology, I would make it lower case consistently.

index.html
```diff
@@ -2630,8 +2630,11 @@ <h3>WoT Security and Privacy Guidelines</h3>
     <section id="sec-servient-implementation" class="informative">
         <h1>Servient Implementation</h1>
         <p>
-            A <a>Servient</a> is a software stack that implements the WoT building blocks presented in the previous section.
-            <a>Servients</a> can host and expose <a>Things</a> and/or consume <a>Things</a> (i.e., host <a>Consumers</a>).
+            As defined in <a href="sec-wot-servient-architecture-high-level"></a>
```
mkovatsc: Needs comma at the end.

index.html
```diff
@@ -2641,61 +2644,64 @@ <h1>Servient Implementation</h1>
...
-            is also a scripting runtime system that is able to interpret and execute application scripts.
-            <a>Servients</a> with the <a>WoT Scripting API</a> usually run on powerful devices, edge nodes, or in the cloud.
-            There are, however, also scripting environments for resource-constrained systems.
-            Furthermore, the WoT Architecture does not limit the application-facing API of the <a>WoT Runtime</a> to JavaScript/ECMAScript.
+            <a href="#architecture-implementation"></a> shows a <a>Servient</a> implementation that is using the (optional) <a>WoT Scripting API</a> building block.
+            Here, the <a>WoT Runtime</a> is also a <a>Scripting Runtime</a> system that,
+            in addition to managing the WoT-specific aspects, also interprets and executes the application scripts.
+            <a>Servients</a> that support the <a>WoT Scripting API</a> usually run on powerful devices, edge nodes, or in the cloud.
+            The WoT Architecture does not limit the application-facing API of the <a>WoT Runtime</a> to JavaScript/ECMAScript [[ECMAScript]].
```
mkovatsc: I think we can drop the reference here, as we DO NOT limit to this, no?

index.html
```diff
@@ -2641,61 +2644,64 @@ <h1>Servient Implementation</h1>
...
+            There are also other runtime systems for resource-constrained systems that can be used to implement a <a>Servient</a>.
```
mkovatsc: This now sounds as if other runtime systems are only for resource-constrained systems... how about:

"Also other runtime systems can be used to implement a <a>Servient</a>."

We have examples and mentioning of resource constraints in the next paragraph, with is the detailed version of the last line.

index.html
```diff
@@ -2641,61 +2644,64 @@ <h1>Servient Implementation</h1>
...
-            for instance C/C++ for embedded <a>Servients</a> or Java for cloud-based <a>Servient</a>.
-            It may also be an alternative scripting langague such as Lua to combine the benefits of application scripts with low resource consumption.
+            for instance C/C++ for embedded <a>Servients</a> or Java for cloud-based <a>Servients</a>.
+            It may also be an alternative scripting languague such as Lua to combine the benefits of application scripts with low resource consumption.
         </p>
         <figure id="architecture-implementation">
             <img src="images/servient/wot-thing-scripting.png"
                  class="wot-arch-diagram" />
-            <figcaption>Implementation of a Servient Using the Optional WoT Scripting API</figcaption>
+            <figcaption>Implementation of a Servient using the WoT Scripting API</figcaption>
```
mkovatsc: You capitalized captions in the Section 5 PR ;)

index.html
```diff
@@ -2641,61 +2644,64 @@ <h1>Servient Implementation</h1>
...
                 (when providing the <a>WoT Scripting API</a> or any other script-based API).
                 They are portable, as they are written against the common <a>WoT Scripting API</a> definitions,
                 and hence can be executed by any <a>Servient</a> featuring this building block.
-                This makes it possible to migrate certain application logic between system components,
+                This makes it possible to shift application logic between system components,
                 for instance moving a <a>Consumer</a> from the cloud to an edge node to meet networking requirements,
-                or move an <a>Intermediary</a> to the cloud to fulfill growing resource demands.
-                The portability can also foster <a>Thing</a> customization by installing behavior following the app concept known from smartphones.
+                or to move an <a>Intermediary</a> to the cloud to fulfill growing resource demands.
+                Additional behavior can be 'installed' after deployment of a <a>Servient</a> through portable applications.
```
mkovatsc: So need portable installer applications? :)

"Portable application scripts also allow to 'install' additional behavior after the deployment of a Servient."

?

index.html
mlagally: This will be resolved by the other PR.

index.html
mlagally: 👍

index.html
mlagally: 👍

index.html
mlagally: Your proposal also suggests that the applications provide the 'install' feature,
however this is a feature of the runtime.

Rewirded to:
Portable applications enable to 'install' additional behavior after the deployment of a Servient.

mlagally:

PR #313, opened 2019-05-07
Cr candidate 2

mlagally:


mlagally:
Updated as discussed in the call on 7.5.
This includes all changes from the four PRs that were reviewed in the call.

PR #319, opened 2019-05-07
Issue 134 device registers itself pattern

mlagally:


mlagally:

PR #338, opened 2019-05-08
[POST-CR-cand3] Align with TD spec terminology and fix 6.7

mkovatsc:
TD spec uses our definition of Thing Descriptions as structured data -> corresponds to TD Info Model.
Serializations are called TD documents.

Section 6.7 had some confusing and (probably hence) partly wrong statements.

Note that this goes from CR-candidate3 to master...


mlagally:
Discussion in arch call on 9.5.:
The  CR "[POST-CR-cand3] Align with TD spec terminology and fix 6.7" was coming in very late, i.e. 10 hours before the arch call and it was not possible to be properly reviewed in the call.
This appears to be an editorial change that does not affect any normative language, so we should defer it to after the CR.

mkovatsc:
This should be closed in favor of #346

PR #340, opened 2019-05-08
Issue 333 minor editorial-arch-call

mlagally:


mlagally:
resolved by merging the combined PR in arch call on 9.5.

PR #343, opened 2019-05-08
Align with TD terminology and fixes for 6.7

mlagally:


mlagally:
@mkovatsc There is a (minor) merge conflict, could be resolved easily. 
Since we are very late in the process, I did not want to interfere with your branch, so please resolve the conflict by yourself.

PR #345, opened 2019-05-09
CR changes since CR-candidate-3 - includes PRs 321, 337 and 340

mlagally:


mlagally:
Approved in arch call on 9.5.

PR #353, opened 2019-05-21
apply feedback from the CR version

ashimura:
I've  checked the diffs between the [master branch version](https://w3c.github.io/wot-architecture/)  and the [CR published version](https://www.w3.org/TR/2019/CR-wot-architecture-20190516/Overview.html) and generated a pullrequest to apply the changes.

Note that some of the edits can't be applied to the master branch version due to ReSpec's restriction, we still need some minor edit when we publish a Proposed Recommendation.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,7 +1,6 @@
 <!DOCTYPE html>
 <html lang="en-US">
 <head>
-<meta http-equiv="content-type" content="text/html; charset=UTF-8" />
 <meta charset="utf-8" />
 <title>Web of Things (WoT) Architecture</title>
 <script class="remove"
@@ -26,6 +25,9 @@
             repoURL: "https://github.com/w3c/wot-architecture",
             branch: "master"
         },
+        implementationReportURI: "https://w3c.github.io/wot-thing-description/testing/report.html",
+        previousPublishDate: "2019-05-16",
+        previousMaturity: "CR",
         editors : [ {
             name : "Matthias Kovatsch",
             w3cid : "75998",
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -273,6 +275,32 @@
...
+    preliminary implementation report</a>.
+    </p>
+
+    <p>This document was published by the <a href=
+    "https://www.w3.org/WoT/WG/">Web of Things Working Group</a> as
+      an Editor's Draft.
+     This document is intended to become a <abbr title="World Wide Web Consortium">W3C</abbr> Recommendation.</p>
+
+    <p>Please see the Working Group's
+    <a href="https://w3c.github.io/vc-test-suite/implementations/">
```
k-toumura: this should be 'https://w3c.github.io/wot-thing-description/testing/report.html'.

mkovatsc:
> https://w3c.github.io/vc-test-suite/implementations/

That is something that should probably also be fixed in the CR at https://www.w3.org/TR/2019/CR-wot-architecture-20190516/

mlagally:
@ashimura I looked at the diff and am not sure I understand your comment about the minor edit for respec. Do you think we can merge this PR after review?

ashimura:
That's true... I'll talk with the Webmaster about that.

index.html
ashimura: fixed.

ashimura: fixed.

ashimura:
@mkovatsc no problem with the CR now :)

index.html
```diff
@@ -612,7 +640,7 @@ <h3>Consumer</h3>
                     and other UX devices.</p>
                 <figure id="smart-home">
                     <img src="images/wot-use-cases/smart-home.png"
-                        class="wot-arch-diagram"/>
+                        class="wot-arch-diagram"/ alt="smart home use case" />
```
mkovatsc: The slash before `alt` has to go.

index.html
```diff
@@ -3238,7 +3266,7 @@ <h3>Connection Through Proxy Synchronization</h3>
                 </p>
                 <figure id="deployment-service-sync-proxy">
                     <img src="images/deployments/arch-deploy-service-sync-intermediary.png"
-                         class="wot-arch-diagram" />
+                         class="wot-arch-diagram" alt="service sync proxy" />
```
mkovatsc: might want to call it "intermediary" also in the alt text.

index.html
ashimura: fixed.

index.html
ashimura: In that case, maybe it would be better to use "deployment-service-sync-intermediary" for id as well. What do you think?

mkovatsc: Yes, sounds good. Need to double check if it is referenced from somewhere and fix there as well...

index.html
mkovatsc: Good, this was the main issue -- approved changes

index.html
ashimura: OK. I've checked the index.html file again for "proxy" and converted it to "intermediary" if appropriate, e.g., the PNG file's name is *-intermediary.png".

Let's review the changes during the Architecture call tomorrow on May 23.

PR #355, opened 2019-05-23
Update index.html

mlagally:
Rename Security Considerations to Security Guidelines, where applicable.


mmccool:
This looks ok to merge.   When this goes to PR though we might want to replace the reference with a dated one (and maybe let ReSpec handle resolving it, rather than having a local reference).   But for now I would go ahead with this patch.

PR #367, opened 2019-06-28
restoring images for TAG explainer

mlagally:


mlagally:
Due to the restructuring of the image directory, the references from the TAG explainer were broken.
Since the TAG explainer is frozen I provided the diagrams under the old name.

PR #368, opened 2019-07-02
Delete definitions of Security and Privacy

mmccool:
NOTE: This PR should only be applied after we have received official TAG feedback and have gotten group consensus.   But it seems likely it will be a recommended change, so I am queuing up a PR for it.

Based on feedback [here](https://github.com/w3ctag/design-reviews/issues/355), remove definitions of the terms "Security" and "Privacy" from the WoT Architecture document.   Note that these terms are, however, defined in the WoT Security and Privacy Guidelines document.

This PR DOES leave the definition for "Personally Identifiable Information" in place.  In a separate issue we may want to find and reference an external definition of this term rather than defining it ourselves.

This PR resolves issue https://github.com/w3c/wot-security/issues/134

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/368.html" title="Last updated on Sep 6, 2019, 7:34 AM UTC (e41264d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/368/44f8dbd...mmccool:e41264d.html" title="Last updated on Sep 6, 2019, 7:34 AM UTC (e41264d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -441,11 +441,6 @@ img.wot-arch-diagram {
             <dd>Information that can be associated with a
                 unique individual.</dd>
             <dt>
-                <dfn>Privacy</dfn>
-            </dt>
-            <dd>The system should maintain the
-                confidentiality of <a>Personally Identifiable Information</a>.</dd>
-            <dt>
                 <dfn>Property</dfn>
             </dt>
             <dd>An Interaction Affordance that exposes state of the Thing.
@@ -458,11 +453,6 @@ img.wot-arch-diagram {
             <dd>The mapping from an Interaction Affordance to concrete messages of a specific protocol,
                 thereby informing Consumers how to activate the Interaction Affordance.
                 W3C WoT serializes Protocol Bindings as hypermedia controls.</dd>
-            <dt>
-                <dfn>Security</dfn>
-            </dt>
-            <dd>The system should preserve its integrity
-                and functionality even when subject to attack.</dd>
             <dt>
                 <dfn>Servient</dfn>
             </dt>


mmccool:
Note that technically definitions are normative but deleting these definitions will not change any implementation, so it *should* be permissible.  Best to check with Kaz first.

takuki:
One source of definition for PII is [IIC Vocabulary Technical Report](https://www.iiconsortium.org/pdf/IIC_Vocab_Technical_Report_2.1.pdf).

The PII in particular is defined [here](https://hub.iiconsortium.org/portal/GlossaryItem/5a04bd421a0dce607c7ffe3c).

mlagally:
A good reference for PII is section 2.9 of https://www.iso.org/obp/ui/#iso:std:iso-iec:29100:ed-1:v1:en

mlagally:
I created a PR with a PII reference, please review:
https://github.com/w3c/wot-architecture/pull/369

mlagally:
After checking the arch document in the call on 4.7. we think it would be better to define the terms "security" and "privacy" in the arch document.
For Privacy the document https://www.iso.org/obp/ui/#iso:std:iso-iec:29100:ed-1:v1:en 
could be used.

mmccool:
It is also fine to define Privacy and Security using external references.   The reason I defined them in the first place is that they are often confused.  However, it is reasonable and correct to say that these should not be normative definitions, just clarifying usage.   It would be odd, however, to only define Privacy using an external reference and not "Security".  So the question is, can we find an external reference for Security as well?   I will bring this up in the security call...

mmccool:
ISO/IEC 27000:2018, section 3.28 has a definition of "information security" which includes availability, integrity, and confidentiality.   This would be a good reference, but I would suggest that we define "Security" as a shorthand: "In this document and in the WoT documents generally, the term Security is to be considered equivalent to the term Information Security defined in...".   "Security" by itself is not actually defined in this ISO document, as I imagine there are conflicts with its use for other forms of security (eg bodyguards, buildings, etc.).   But in our context I think it is reasonable to define the short form.

mmccool:
It's interesting the IIC document cited above includes "privacy" and terms like "information security risk" but oddly not "information security" or "security" (although it does have "physical security").  I assume this is just an oversight...

mmccool:
During the security call on July 8 we agreed that modifying the definitions to refer to the ISO standards makes sense, but defining "Security" to be equivalent to "Information Security" is also a pragmatic approach.   We also agreed to add a definition (referencing ISO...) for "Physical Security" to distinguish it from "Information Security".   I will update this PR to reinstate the definitions with these changes, and so have changed the title from "Delete" to "Modify".

mlagally:
> It's interesting the [IIC Vocabulary document cited above](https://www.iiconsortium.org/pdf/IIC_Vocab_Technical_Report_2.1.pdf) includes "privacy" and terms like "information security risk" but oddly not "information security" or "security" (although it does have "physical security"). I assume this is just an oversight...

There *is* a definition for "security" in the IIC document, which we could use. Note that it contains basically the same text as the definition of "Information security" in https://www.iso.org/obp/ui/#iso:std:iso-iec:27000:ed-5:v1:en

mmccool:
Sorry I did not yet get around to updating this PR to include "standard" definition with references for Security and Privacy.   That is my intention, and I plan to get it done today in time to be reviewed in the Architecture call tomorrow.

mmccool:
For technical reasons, I am closing this (without a merge!) and replacing it with PR https://github.com/w3c/wot-architecture/pull/384.   The new PR adds definitions based on ISO standards rather than deleting them as this PR does.

PR #378, opened 2019-07-12
fix broken reference for "personally identifiable information"

mlagally:


index.html
```diff
@@ -178,7 +178,7 @@
                 status: "Recommendation",
                 date : "January 2015"
             },
-	    "ISO/IEC 29100" : {	
+	    "ISO/IEC-29100" : {
```
k-toumura: We can't use a slash (`/`) for reference ID. 
RegEx of reference ID is `"^[a-zA-Z0-9\\-_+\\.]+$"` (from [schema for SpecRef](https://github.com/tobie/specref/blob/master/schemas/raw-reference.json)).

index.html
```diff
@@ -453,8 +453,8 @@ <h1>Terminology</h1>
                 <dfn data-lt="Personally Identifiable Information">Personally Identifiable Information (PII)</dfn>
             </dt>
             <dd>Any information that can be used to identify the natural person to whom such information relates, 
-		or is or might be directly or indirectly linked to a natural person, as defined in section 2.9 of
-		[[!ISO/IEC 29100]].
+		or is or might be directly or indirectly linked to a natural person. Note that we use the same 
+		definition as [[ISO/IEC-29100]].
```
k-toumura: '</dd>' tag is needed at the end of this sentence.

Issue #371, opened 2019-07-04
TAG interoperability concerns


mlagally:
David Baron:
I'm also concerned that the architecture overview says "The format can be processed either through classic JSON libraries or a JSON-LD processor": is this defined in a way such that both types of processing will yield the same results in all cases? If not, what leads to interoperability?

takuki:
Note: This is with regards to a TAG comment https://github.com/w3ctag/design-reviews/issues/355#issuecomment-505228840 from @dbaron.

takuki:
I created PR #379 for this issue.
Please comment and make suggestion or improvements to the PR.

dbaron:
I was asked to give some examples of interoperability concerns.  Here are two, one in each direction:

1. A developer is asked to optimize the file size of thing descriptions.  Since this developer thinks that using content negotiation to serve either a compressed or uncompressed version wouldn't address the reason there was concern over file sizes to begin with, she decides to use her knowledge of JSON-LD, make her own copy of the context document https://www.w3.org/2019/wot/td/v1, rename a bunch of the JSON property names to be shorter.  For example, she modifies this bit: `"securityDefinitions":{"@id":"td:securityDefinitions","@type":"@id","@container":"@index"}` to instead be `"sd":{"@id":"td:securityDefinitions","@type":"@id","@container":"@index"}` (and a bunch of similar changes), and then changes all of the thing descriptions in her system to reference this modified context document instead of the standard one, and also to use the `sd` name instead of `securityDefinitions` (and other similar changes).  She tests that this works against a bunch of implementations, and it works fine since all the implementations she tests process the thing descriptions as JSON-LD (is there a correct term for this?), and then deploys the change.  This change, however, reduces interoperability since it breaks the use of these thing descriptions in any implementation that processes them as JSON.

2. A developer (in an project unrelated to the one in the first example) receives a bug report that processing thing descriptions is sometimes slow due to the network latency needed to retrieve https://www.w3.org/2019/wot/td/v1 .  (Or maybe, instead, this was the same bug report the previous developer got, about optimizing file size.)  She looks at the thing description files and determines that the `@context` bit isn't actually needed.  She tests her change in a bunch of implementations, and it works fine, since all of the implementations she tests process the thing descriptions as JSON (rather than as JSON-LD, as in the previous example).  So she deploys the change.  This change reduces interoperability in the opposite direction:  it breaks the use of these thing descriptions in implementations that process them as JSON-LD.

[...]

PR #379, opened 2019-07-15
Described what JSON-LD additionally brings to processing (Issue #371)

takuki:
Note: This is with regards to a TAG comment https://github.com/w3ctag/design-reviews/issues/355#issuecomment-505228840 from @dbaron. 

This PR, in particular is, about clarifying the difference between JSON and JSON-LD processing of a TD.


mlagally:
@takuki Thanks, this looks good to me, but let's wait for @vcharpenay for his comments.

PR #396, opened 2019-10-11
Update index.html - Add security data/metadata defns

mmccool:
NOTE: all changes here included in https://github.com/w3c/wot-architecture/pull/400

Add definitions for the following terms to the Terminology section:
- Private Security Data
- Public Security Metadata
Also add the following terms used by the above:
- Metadata
- Security Configuration
Also various changes through the document to mark up these terms where they appear to link back to the definition.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/396.html" title="Last updated on Oct 17, 2019, 5:28 AM UTC (534c223)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/396/4a1fcb0...mmccool:534c223.html" title="Last updated on Oct 17, 2019, 5:28 AM UTC (534c223)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -469,8 +469,7 @@ img.wot-arch-diagram {
             <dt>
                 <dfn>Metadata</dfn>
             </dt>
-            <dd>Data that provides a description of an entity's abstract characteristics or situtation
-                as opposed to concrete measurements.
+            <dd>Data that provides a description of an entity's abstract characteristics.
                 For example, a <a>Thing Description</a> is Metadata for a <a>Thing</a>.</dd>
             <dt>
                 <dfn data-lt="Personally Identifiable Information">Personally Identifiable Information (PII)</dfn>
@@ -534,7 +533,7 @@ img.wot-arch-diagram {
             <dt>
                 <dfn>Security Configuration</dfn>
             </dt>
-            <dd>The combination of Public Security Data, Private Security Metadata, and any other configuration
+            <dd>The combination of Public Security Metadata, Private Security Data, and any other configuration
                 information (such as public keys) necessary to operationally configure the security mechanisms of a Thing.</dd>
             <dt>
                 <dfn>Servient</dfn>
--- a/index.html
+++ b/index.html
@@ -467,6 +467,12 @@ img.wot-arch-diagram {
                 application-facing APIs, data model, and protocols or
                 protocol configurations.</dd>
             <dt>
+                <dfn>Metadata</dfn>
+            </dt>
+            <dd>Data that provides a description of an entity's abstract characteristics or situtation
+                as opposed to concrete measurements.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -506,6 +531,11 @@ <h1>Terminology</h1>
                 We additionally note that it is desirable that these properties be maintained both in normal operation
                 and when the system is subject to attack.
             </dd>
+            <dt>
+                <dfn>Security Configuration</dfn>
+            </dt>
+            <dd>The combination of Public Security Data, Private Security Metadata, and any other configuration
```
mlagally: Reverse Data and Metadata

index.html
```diff
@@ -466,6 +466,12 @@ <h1>Terminology</h1>
                 Mozilla Project Things with its own specifications for
                 application-facing APIs, data model, and protocols or
                 protocol configurations.</dd>
+            <dt>
+                <dfn>Metadata</dfn>
+            </dt>
+            <dd>Data that provides a description of an entity's abstract characteristics or situtation
```
mlagally: typo

mlagally: suggest to just use:
Data that provides a description of an entity's abstract characteristics.

index.html
mmccool: fixed

index.html
mmccool: fixed.

mmccool:
issues noted fixed, ready to merge.

index.html
mmccool: fixed

mmccool:
One thing I have been thinking about: I refer to "public security data" in "Security Configurations" but don't have a definition for it to parallel "Private Security Data" and "Public Security Metadata".  Basically this would include the public keys, etc, typically encoded in a certificate, but for various reasons we DON'T want to include that in the TD.   However... for consistency and completeness, should I add a definition for "Public Security Data" so that the "Security Configuration" can refer consistently to all three?

mlagally:
> One thing I have been thinking about: I refer to "public security data" in "Security Configurations" but don't have a definition for it to parallel "Private Security Data" and "Public Security Metadata". Basically this would include the public keys, etc, typically encoded in a certificate, but for various reasons we DON'T want to include that in the TD. However... for consistency and completeness, should I add a definition for "Public Security Data" so that the "Security Configuration" can refer consistently to all three?

The current specification is consistently using these terms (public security metadata, provate security data), also in the pictures.
I think that introducing an additional definition of public security data would not help to make things clearer, since the border between data and metadata depends on the deployment.

mmccool:
So I wanted to address the point raised by PLH regarding discussing the ID risk for TDs, but unfortunately there is a risk of merge conflicts because of the changes made to terminology in this PR.   therefore, in the interests of time, I'm going to try to address that issue in this PR as well.

PR #397, opened 2019-10-14
Refactoring and rewriting abstract and introduction

mmccool:
A refactoring and rewrite of the abstract and introduction to execute the changes discussed for the next CR, specifically to position this specification as an abstract architecture.   Changes include:
1. Moving material describing the building blocks from the abstract to the introduction.   This significantly shortens the abstract (which was getting too long) and lets us more clearly state the purpose of the specification: to provide an *abstract* architecture.  
2. Removing a paragraph that over-emphasized mechanisms included in other documents.  This paragraph was a little confusing even though it was talking about the building blocks because of where it was positioned in the abstract (second paragraph).  I replaced it with a single sentence in the abstract that mentioned the building blocks without going into detail... the details are in the introduction.
3. Rewording the discussion of the Security and Privacy building blocks (now in the introduction) to emphasize that a concrete set of mechanisms and an implementation are needed to evaluate security and privacy, and that this document and the WoT generally does NOT fully specify such an implementation... and when used "descriptively", cannot... but that nevertheless we provide some general high-level guidelines.
4. Rewording various things so that the introduction flows better.
5. Removed certain phrases like "so-called" that have negative connotations in English.   This means in particular that I reworded the description of Binding Templates/Protocol Bindings.

Note: in the end this ended up being more of a rewrite than a patch, so I suggest you don't bother with the diff and just look at the rendered version; also, read both the abstract and the introduction.

No other sections than the abstract and introduction were modified.

### DIFF
--- a/index.html
+++ b/index.html
@@ -216,67 +216,39 @@ img.wot-arch-diagram {
         <p>The W3C Web of Things (WoT) was created to enable
             interoperability across IoT platforms and application
             domains.</p>
+
+<!-- This paragraph is a little inconsistent since we talk about "mechanisms" but
+     then immediately talk about this being an "abstract architecture" that does
+     NOT provide mechanisms.  I suggest we leave it out to avoid confusion. 
+     Its content is basically covered by the discussion of building blocks in
+     the Introduction.  However, if we take it out, a sentence mentioning
+     "building blocks" is useful so I added one to the next paragraph.
+
         <p>WoT provides mechanisms to formally describe IoT
             interfaces to allow IoT devices and services to communicate
             with each other, independent of their underlying
             implementation, and across multiple networking protocols. In
             addition, WoT offers a standardized way to define and program
             IoT behavior.</p>
+-->
 
         <p>This <em>WoT Architecture</em> specification describes the abstract
-            architecture for the W3C Web of Things. It is derived from a
+            architecture for the W3C Web of Things. 
+            This abstract architecture is based on a
             set of requirements that were derived from use cases for
-            multiple application domains. 
-	    Since the Web of Things is addressing a new area,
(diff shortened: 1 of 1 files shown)


mlagally:
Great work and really an improvement of the previous introduction.

There seems to be missing text describing the scripting API - I believe this has been acidentially deleted. Can you please take another look?

mmccool:
OK, I will look into the accidentally-deleted-text issue and do an update shortly.

mmccool:
I'm not sure what you think is missing.   I checked and the description of the Scripting API is still there.  What I DID delete (on purpose) was the second paragraph of the abstract which said

"WoT provides mechanisms to formally describe IoT interfaces to allow IoT devices and services to communicate with each other, independent of their underlying implementation, and across multiple networking protocols. In addition, WoT offers a standardized way to define and program IoT behavior."

My problem with this paragraph was the "WoT provides mechanisms" part and its position in the very second paragraph.   However, deleting this paragraph also deleted the last sentence, which talks about the scripting API...

Assuming this is what you are talking about, I can work this material into the introduction but I'll have to soften the "WoT provides mechanisms" part.

mmccool:
OK, I put those two sentences back in, but reworded them to say "WoT supports" rather than "WoT provides mechanisms" (I just want to avoid the word "mechanism").  It's actually not incorrect to say that the WoT TD provides a mechanism for describing the network interface of IoT devices, since it DOES, it's just that the previous location of this statement clashed a bit with the description of the purpose of the WoT Architecture document, which explicitly does NOT provide mechanisms.

Please confirm that this is the "missing content" you were talking about.   Note I think it scanned fine without these sentences since they are just repeating things that are said in other words in the descriptions of the building blocks.

PR #413, opened 2019-11-25
Moving text and figure on lifecycle from security doc to architecture

ereshetova:
As we discussed last week during security call, we need to start the work to move the lifecycle text from the security document to the architecture document, where it belongs. 

Feel free to merge this and adjust/rewrite this as needed, we will adjust the security doc to be aligned with lifecycle defined in the architecture doc.

***
<a href="https://pr-preview.s3.amazonaws.com/ereshetova/wot-architecture/pull/413.html" title="Last updated on Nov 25, 2019, 10:37 AM UTC (2436b65)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/413/a43bc6d...ereshetova:2436b65.html" title="Last updated on Nov 25, 2019, 10:37 AM UTC (2436b65)">Diff</a>


ereshetova:
@mmccool please take a look

mlagally:
The proposed MR puts the lifecycle into the section 6 which is normative.
Since we cannot enforce a specific lifecycle model on devices (specifically on legacy devices) we need to discuss if this is the right place.

ereshetova:
The location was a mere guess from my side, I just had to start somewhere. 
The main thing is that architecture document should be defining the lifecycle and security document just refers to it and defines needed notions on top.

ereshetova:
Per decision from architecture call on 5.12, I will resubmit this to a separate 1.1 branch

PR #428, opened 2020-01-23
Digital Twin Use Case + editorial cleanup

mlagally:
This PR contains the use case for digital twins.
In addition it contains some cleanup, fixing typos and adding titles to the mmi use cases.


mlagally:
reviewed in arch calls on Feb 13th

PR #431, opened 2020-02-06
Create X-Protocol Interworking.md

mlagally:


USE-CASES/X-Protocol Interworking.md
```diff
@@ -0,0 +1,89 @@
...
+Device owners, cloud providers.
+
+
+<List all users that are involved in the use case, e.g. device manufacturer, gateway manufacturer, cloud provider>
+
+### Motivation:
+
+<Provide a description of the problem that is solved by the use case and a reason why this use case is important for the users>
+
+In home and industrial scenarios various devices are connected to a common network. These devices implement different protocols. To enable interoperability, an "agent" needs to communcate across different protocols. Platforms for this agent can be  edge devices, gateways or cloud services.
```
egekorkan: typo at communicate

USE-CASES/X-Protocol Interworking.md
```diff
@@ -0,0 +1,89 @@
...
+
+Interoperability across protocols is a must for all user scenarios that inctegrate devices from more than one protocol.
```
egekorkan: typo at integrate

mlagally:
On 6. Feb 2020, at 10:53, Ege Korkan <ege.korkan@tum.de> wrote:
 
Hi Michael(s),
A way to bridge the protocols is important. I would start by defining that any protocol that needs to bridge all other ones, MUST have a mapping for ALL the op values defined in the TD and Architecture documents. For example, we cannot use MQTT as the one protocol to bridge everything since it doesn't have proper readproperty support and impossible to do readmultipleproperties (these would be addressed if you use this new response topic feature of MQTT5). 
I would imagine some websockets based solution would hurt the least amount of people and be compatible with the most use cases.
Best,
Ege

mlagally:
Hi Ege, 
 
thanks a lot for your quick turnaround and for flagging the MQTT problem.
 
I think it will be difficult if we have to say that WoT can be interoperable but not for MQTT.
 
The MQTT case has two angles: 
If a MQTT exposed device does not support certain OPs, the TD just does not offer them. 
If a MQTT consumer wants to read a value from another device and does not have a way to do a direct read operation,
the TD of the device has to provide another way (e.g. subscribe or action) for querying a property.
If it does not offer this way, the device is not interoperable.
 
This is a requirement / constraint we should formally describe in the profile spec.
  
Thanks,
 
Michael

mlagally:
On Feb 6, 2020, at 5:36 AM, Mccool, Michael <michael.mccool@intel.com> wrote:

One thing we should discuss is whether bridging “any” protocol is a design goal of WoT, or if the idea is just to expose the operations that a Thing’s protocol supports.  I think the former is the “proxy” pattern and may be a desirable goal, but may require a “mapping” process between protocols that don’t have compatible operations (as a simple example, supporting a protocol that does not support readmultiple could simply see the proxy issue multiple reads to implement it).
 
I suppose I should comment on this in the issue tracker rather than on an email thread that is likely to get lost…
 
Michael M

mlagally:
On 6. February 2020 at 15:26:29 CET Michael Koster <michael.koster@smartthings.com> wrote:

Thanks for bringing this up again. I have thought some of this through a little further.

The whole point of the proxy pattern for bridging is that is doesn’t require a mapping from one protocol to the other. This is done through the WoT meta-operations (ReadProperty, WriteProperty, InvokeAction, subscribeEvent, …). If protocol A and protocol B bpth support a given meta-operation, then each protocol only needs to map to the WoT meta operations, and the translation is done in two steps, to- and from-. (or one step through some logic minimization/code generation technique.

A couple of questions come up in the case that a protocol doesn’t support the semantics of a given meta-operation

0. Are the semantics of the WoT meta-operations well enough defined?

1. Is support of all meta-operations required for interoperability? 

2. Are there other meta-operations that should be added?

For example, consider the cloud-device interaction pattern that is popular with IoT services:

1. Device “registers” and uploads it’s metadata 
2. Device uploads its state in one big dump
3. Device sends compact updates whenever something changesor events occur
4. Service sends compacts commands and updates to the device to affect the state

- There isn’t a ReadProperty operation. The service needs to provide this from its cached information
- WriteProperty and InvokeAction use the same communication pattern 
- SubscribeEvent and ObserveProperty use the same commnication pattern

How do we structure the protocol binding for this case? We can describe the communication, but what does the application expect?

Looking forward to today’s discussion.

Best regards,

Michael

mlagally:
@mjkoster 
The cloud-device pattern causes some head-ache, since it reverses the roles: 

The cloud service is actually in the device role (implementing a REST interface to a digital twin) and the thing is a consumer.

From a scaling perspective this is the right model, since it only initiates network traffic when some relevant device state has changed. The "thing implements server" approach leads to overhead, (e.g. managing event subscriptions and timeouts, potentially keeping sockets alive etc.) This does not scale well.

USE-CASES/X-Protocol Interworking.md
mlagally: fixed

USE-CASES/X-Protocol Interworking.md
mlagally: fixed

mlagally:
Arch call on 13.2.:
We should break this down into categories for smart home, smart cities, smart factories, ...
We also need to determine business value of concrete protocol interworking.

zolkis:
Also, we need to check how does this relate to native bridging specifications for some of the protocols (like OCF).

PR #447, opened 2020-03-04
Create retail.md

mlagally:
Retail use case jointly created by David Ezell and Michael Lagally.


mlagally:
Reviewed in Arch call on March 5th.

PR #470, opened 2020-03-25
Add Transportation use case

zolkis:
A first attempt to frame transportation use cases.


USE-CASES/transportation.md
```diff
@@ -0,0 +1,88 @@
...
+
+#### Variants:
+
+There will be different verticals, such as:
+
+* Smart City public transport
+* Smart City traffic management
+* Smart city vehicle management
+* Cargo traffic management
+* Cargo vehicle management
```
mlagally, line 76: Cargo is ship, air, road, rail and last mile
Stakeholders are very different companies, however shared identifiers for packages, source and target location are desirable, shipping options, like dangerous goods, priority

USE-CASES/transportation.md
```diff
@@ -0,0 +1,88 @@
...
+
+Transportation system implementers will be able to use a unified data description model accoss various systems.
+
+#### Variants:
+
+There will be different verticals, such as:
+
+* Smart City public transport
+* Smart City traffic management
+* Smart city vehicle management
```
mlagally, line 74: e.g. limited traffic zones,

mlagally:
We need wider review from multiple stakeholders, please help.

PR #471, opened 2020-03-26
Create a use case of agriculture

mryuichi:


USE-CASES/smart-agriculture.md
```diff
@@ -0,0 +1,55 @@
...
+
+Agricultural corporation, Farmer, Manufacturers (Sensor, other facilities), Cloud provider
+
+### Motivation:
+
+Greenhouse Horticulture controlled by computers can create an optimal environment for growing plants. This enables to improve productivity and ensure stable vegetable production throughout the year, independent of the weather. This is the result of research on the growth of plants in the 1980s. For example, in tomatoes, switching to hydroponics and optimizing the temperature, humidity and CO2 concentration required for photosynthesis resulted in a five times increase in yield. The growth conditions for other vegetables also have been investigated, and this control system is applied now.
+
+### Expected Devices:
+
+Sensors ( temperature, humidity, brightness, UV brightness, air pressure, and CO2)
```
relu91: Other sensors and devices that we are employing in our agriculture project are: 
- Multispectral cameras that are used to extract the Leaf Area Index parameter of a crop
- Drones that survey the area. Also, they might be used as actuators (i.e. spray pesticide ) 
- Remote-controlled sprinklers
- flow meters to control the amount of water fed to the crop
- Soil moisture, conductivity and temperature
- Automatic weirs to control the flow in agriculture channels and canals
- Water level sensors (again used for channels and canals)

USE-CASES/smart-agriculture.md
```diff
@@ -0,0 +1,55 @@
...
+To maximize photosynthesis, the temperature, CO2 concentration, and humidity in the greenhouse are mainly controlled. When the sunlight comes in the morning and CO2 concentration inside decreases, the application turns on the CO2 generator to keep over 400 ppm, the same as the air outside. The temperature in the greenhouse is adjusted by controlling the heater and the sunlight shielding sheet.
+The cloud gathers all sensor data and the status of the facilities. The application makes the best configuration for the region of the greenhouse located.
+
+#### Variants:
+
+<Describe possible use case variants, if applicable>
+
+### Gaps:
+
+In the case of the wireless connection to the sensors, the gateway should keep the latest value of the sensors since the wireless connection is sometimes broken. The gateway can create a virtual entity corresponding to the sensor and allow the application to access this virtual entity having the actual sensor status like sleeping.
```
relu91: This is very true. We have experienced the same in our pilots. It is not only a problem of intermittent connection but sometimes it is a by-design condition. For example, we are employing low power sensors connected with LoRaWan that sleeps for 1 hour and then sends a data packet. In this case, we are using the "virtual entity" created by the Chirpstack server (i.e. the gateway) to interact with the sensors continuously using MQTT protocol.

relu91:
Hello, I added some comments based on my experience in the [SWAMP project](http://swamp-project.org/). I hope they are useful.

USE-CASES/smart-agriculture.md
mryuichi: @relu91, thank you for your useful comments and very sorry for my late reply. 
I agree many sensors and actuators are used in the agricultural fields. However, there are several types of ICT controlled agriculture like plant factories, Greenhouses, and open field cultivations. They have different requirements for sensors and actuators.
This use case description is for greenhouse horticulture. so, how about adding some of them you mentioned used for the greenhouse to this description?

I think the following sensors are for them;
- Multispectral camera
- Soil moisture, conductivity and temperature

Other sensors and actuators could be used mainly for open fields. Creating another use case for agriculture might be better. What do you think?

USE-CASES/smart-agriculture.md
mryuichi: As you mentioned, this is sometimes by-design condition. What I described in this use case is the necessity of the virtual entity on the gateway for maintaining the status of the IoT device such as sensor instead of the actual device.
Our implementation is very similar to yours. The sensors are connected with LoRa or Bluetooth, which are depending on the distance between sensors and the gateway. They can work 24 hours without power supply since they have  a solar panel and a capacitance

mryuichi:
@relu91 , many thanks for your useful information. I saw some project pages of SWAMP and found the use cases to be for open field cultivations, The requirements derived from your use cases seem different from my use case for greenhouses. So I recommend you to create another user case.

relu91:
Yes, indeed! Sorry, I thought that this was a broader use-case focused on agriculture in general. I actually did not read the title of the document but only the PR title which misguided me. 

I guess that some requirements are common between the use cases, but I agree to create a new Use case specifically for open field cultivations. Thanks for your guidance!

zolkis:
Like with transportation, this is too big a topic and as discussed in the last architecture call, it would make sense to create a _category_ for agriculture (a directory with a readme.md as index and use cases as separate md files).

egekorkan:
Blocked previous spam commenter

PR #483, opened 2020-04-22
Folder cleanup - removing obsolete diagrams

mlagally:
This is a housekeeping MR.

It is removing the obsolete USE-CASES/WoT lifecycle diagram-WoT new lifecycle.svg diagram, which is 2 months old and has been superseeded by the more recent https://github.com/w3c/wot-architecture/blob/master/proposals/WoT%20lifecycle%20diagram-WoT%20new%20lifecycle.svg

Additionally it removes the svg version of the unified lifecycle diagram and preserves only the Draw.io version to avoid double-booking: https://github.com/w3c/wot-architecture/blob/master/proposals/unified%20device%20lifecycle.drawio


mlagally:
Manually cleaned up

PR #498, opened 2020-05-07
Adding profile use case

mlagally:


USE-CASES/wot-profile.md
```diff
@@ -0,0 +1,90 @@
+## Title: <Pick a descriptive title>
```
mlagally, line 1: Title: Out of the box interoperability

PR #500, opened 2020-05-11
Update retail.md with Security and Privacy considerations

mmccool:
Merging S&P considerations from Clerley and Conexxus.  Still needs further discussion before merging.

### DIFF
--- a/USE-CASES/retail.md
+++ b/USE-CASES/retail.md
@@ -69,6 +69,29 @@ by providing real-time visibility into worker status, location, and work environ
 
 - Use edge computing, in particular video analytics, in combination with IoT devices to deliver an enhanced 
   customer experience, better manage inventory, or otherwise improve the store workflow.
+  
+### Security Considerations:
+ 
+- Use the latest known industry standards for securing data at rest and over communication.
+  Always encrypt data.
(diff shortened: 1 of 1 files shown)


mmccool:
David had the following comments (paraphrased)
1. Can we add some references?
2. What about replay attacks, e.g. from the device to the cloud?   Timestamps should be added to every request.  (Comment from McCool: this is one possible approach to prevent replay; there may be others, for example, block chains.  Also note that small devices may not have actual clocks, so this may be a "sequence number".   It also needs to be protected, e.g. by including it in a signed hash.)
3. What about DoS attacks (on cloud services, local services, devices?)

mmccool:
The proposed considerations are “typical” high-level considerations for security (which we definitely need to check that we include in our Best Practices document) but did not seem very specific to retail.  I’m wondering if there are more specific considerations for the retail context; for example, preserving the privacy of customers, the security of monetary transactions, detection of modification of security services, and the like.   The last one overlaps with “tamper-proof”, but I think tamper-proof is a requirement that should be derived from a use-case saying something like “Do not allow customers and unauthorized sales representatives from modifying, reconfiguring, or disabling devices, and notifying management if they do attempt this.”

PR #502, opened 2020-05-14
Add education and shared resources use case

egekorkan:


USE-CASES/education.md
```diff
@@ -0,0 +1,90 @@
...
+### Privacy Considerations:
+
+none from the WoT point of view since we want the devices to be used by anyone and the devices do not share any information that is related to the students or us as the provider of the devices.
+However, there are cameras which can show humans entering the room as a side effect (they are meant to monitor the devices). The streams are accessible only to authorized users, the room has signs on the door and there is a cage around the area that is filmed.
+
+### Gaps:
+
+#### Thing Description
+
+- How to give hints that a particular action should not be used by others at the same time. A new keyword (like `"shared":true`) would be needed for devices that do not implement a describable mechanism.
```
relu91, line 80: That is almost the same problem that I introduced in https://github.com/w3c/wot-thing-description/issues/899#issuecomment-627836590. In particular here:

> We are leaving out the problem of ownership, what happens if some other consumer tries to cancel an action started by me. However, it might be out of scope.

 I am glad that it is described in a proper use case! Great work Ege.

PR #505, opened 2020-05-20
Create a new requirements from agriculture

mryuichi:
Create a draft of requirements derived from the agriculture use case.
There are three types of requirements in the description.


mlagally:
Arch call on 28.5. 
People would like more time for review, name change was suggested.

mlagally:
Withdrawn in Arch call on 3.9. by Marsukura-san.

PR #509, opened 2020-05-28
Add smart building use case

sebastiankb:
Provide smart building use case


mlagally:
Arch call on 28.5.
Please consider adding Discovery if it is part of the use case.

PR #516, opened 2020-06-01
Created smart-campus use case

AndreaCimminoArriaga:
Use case focused on discovery capabilities within a smart campus. This UC was originally proposed in the wot-discovery repository, but I was asked to move it here.


mlagally:
Arch call on 18.6.:
Very good submission with a lot of details, thanks a lot.
The images should be moved to the images directory of this github repo.
If possible, the images should be available as SVG files to have them better rendered in the document.

AndreaCimminoArriaga:
Many thanks! I included images in svg and png

PR #531, opened 2020-08-26
Define Discovery requirements

mmccool:
To support the WoT Discovery deliverable.


mlagally:
Approved in Arch call on 27.8.

PR #562, opened 2020-10-14
Add candidates for UI link types

relu91:
Coming from https://github.com/w3c/wot-testing/issues/87, this PR adds two possible link types to point towards WebThing UIs. See [my comment](https://github.com/w3c/wot-testing/issues/87#issuecomment-702801947)

### DIFF
--- a/REQUIREMENTS/link-relation-types.md
+++ b/REQUIREMENTS/link-relation-types.md
@@ -44,6 +44,7 @@ when you think about how you model things in UML: inheritance, implementation, a
 - "Previous": version of the document
 - "member", "Item"
 - "manifest": point to SW implementing the TD
+- "alternate": point to other rappresenatations of the Web Thing. (i.e. HTML document, RDF ...)
 - "Profile": denote compliance
 
 #### Not in IANA list:
@@ -51,6 +52,7 @@ when you think about how you model things in UML: inheritance, implementation, a
 - "twin-of"
 - "shadow": caching data
 - "simulation", could be a shadow or a twin
+- "application", refers to a companion application
 
 ### Entity types:
 * Thing model


REQUIREMENTS/link-relation-types.md
```diff
@@ -44,13 +44,15 @@ when you think about how you model things in UML: inheritance, implementation, a
 - "Previous": version of the document
```
danielpeintner, line 44: I think this should be lower-case
according to https://www.iana.org/assignments/link-relations/link-relations.xhtml

REQUIREMENTS/link-relation-types.md
```diff
@@ -44,13 +44,15 @@ when you think about how you model things in UML: inheritance, implementation, a
...
 - "member", "Item"
 - "manifest": point to SW implementing the TD
+- "alternate": point to other rappresenatations of the Web Thing. (i.e. HTML document, RDF ...)
 - "Profile": denote compliance
```
danielpeintner, line 48: "profile" -> lower case also

relu91: yeah, I found the notation a bit strange. Plus some of them are not even IANA, but reading the text it seems they are (there is a dedicated section for NO-IANA proposal). I think the document should be reviewed, I can do that in this PR. Just noticed that I did not modify what you're comment is pointing to, it was there before my changes.

REQUIREMENTS/link-relation-types.md
relu91: See my other [comment](https://github.com/w3c/wot-architecture/pull/562#discussion_r505625825).

REQUIREMENTS/link-relation-types.md
danielpeintner: Even though I know that @relu91 did not add the initial capitalized terms but I think we can change them in this PR ..

mlagally:
Reviewed in arch call on Nov 26.
The concrete list of types need to be refined, however the current candidates are a good starting point.

PR #577, opened 2021-01-18
Introduce Partial TD definition

relu91:
I tried to simplify the definition given during the last call. In particular, I dropped the term _subset of_ because it could be understood as if a Partial TD could contain any element of the TD model without any particular order or structure (much like a TD fragment). I also remove the information about the serialization because I found it redundant.  

@ashimura I linked the scripting API document in the _Note_. Did I use the correct link? Should I refer to the published document or the TR is ok? Sorry, I'm still confused a little bit about the publication process.

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-architecture/pull/577.html" title="Last updated on Jan 19, 2021, 3:14 PM UTC (b939bd5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/577/591eeb0...relu91:b939bd5.html" title="Last updated on Jan 19, 2021, 3:14 PM UTC (b939bd5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -611,8 +611,8 @@
         A <a>Partial TD</a> is an object that follows the same hierarchical structure of the <a>TD</a> information model, 
         but it may not contain all the mandatory elements. 
         <p class="note">
-          A <a>Partial TD</a> is used in <a>WoT Scripting API</a> as an input for the creation of exposed things. 
-          See the <a href="https://www.w3.org/TR/wot-scripting-api/#the-produce-method">produce</a> method for more details.
+          A <a>Partial TD</a> is used in <a>WoT Scripting API</a> as an input for the creation of <a>Exposed Things</a>. 
+          See the <a href="https://www.w3.org/TR/2020/NOTE-wot-scripting-api-20201124/#the-produce-method">produce</a> method for more details.
         </p>
       </dd>
       <dt>
--- a/index.html
+++ b/index.html
@@ -604,6 +604,18 @@
         shared for an entire group of <a>Things</a>. Compared to a Thing Description, a Thing Model does not contain enough 
         information to identify or interact with a Thing instance.</dd>
       <dt>
+      <dt>
+        <dfn>Partial TD</dfn>
+      </dt>
+      <dd>
+        A <a>Partial TD</a> is an object that follows the same hierarchical structure of the <a>TD</a> information model, 
+        but it may not contain all the mandatory elements. 
+        <p class="note">
+          A <a>Partial TD</a> is used in <a>WoT Scripting API</a> as an input for the creation of exposed things. 
+          See the <a href="https://www.w3.org/TR/wot-scripting-api/#the-produce-method">produce</a> method for more details.
+        </p>
+      </dd>
+      <dt>
         <dfn>Transfer Protocol</dfn>
       </dt>
       <dd>The underlying, standardized application layer


index.html
```diff
@@ -603,6 +603,18 @@ <h1>Terminology</h1>
         capabilities. It describes the <a>Properties</a>, <a>Actions</a>, and <a>Events</a> and common metadata that are 
         shared for an entire group of <a>Things</a>. Compared to a Thing Description, a Thing Model does not contain enough 
         information to identify or interact with a Thing instance.</dd>
+      <dt>
+      <dt>
+        <dfn>Partial TD</dfn>
+      </dt>
+      <dd>
+        A <a>Partial TD</a> is an object that follows the same hierarchical structure of the <a>TD</a> information model, 
+        but it may not contain all the mandatory elements.
```
danielpeintner, line 612: The proposal looks good to me.

An alternative to
"but it may not contain all the mandatory elements. "
may be
"but it does not have to contain all mandatory elements."

I leave it to native speakers what is more clear....

ashimura:
@relu91 thanks for generating this PR!

Regarding the question of "which document to be cited for the reference", my suggestion is (always) using the dated "This version" URL of the published TR, i.e., https://www.w3.org/TR/2020/NOTE-wot-scripting-api-20201124/#the-produce-method, because  the content of the undated "Latest published version URL (I.e., https://www.w3.org/TR/wot-scripting-api/#the-produce-method) tend to get changed.

relu91:
@ashimura  Thank you for your guidance.  I've updated the PR accordingly.

relu91:
Another point, should we define the full name? In other words, should I rename the <dfn> element to _Partial Thing Description_ ?

danielpeintner:
> Another point, should we define the full name? In other words, should I rename the element to _Partial Thing Description_ ?

we used "partial TD" in the past and I think it is fine like it is without expanding it....
Anyhow, I don't have a strong opinion.

danielpeintner:
FYI: In today's scripting call we discussed the possibility of removing any mentioning of "partial TD" from the architecture terminology section.

Let me describe our thinking. As we see it, "partial TDs" are solely used in scripting (OR are there other documents referring to it?) and the scripting document could define it if needed. Having said that, in scripting we are thinking to use the term "ExposedThingInit" which seems more aligned with the web platform terminology...

We (@relu91, @danielpeintner, @zolkis, ...) are happy to also discuss it further in the architecture call.

zolkis:
For a larger context, the TD TF defines the outcome (Thing Description) for an exposed Thing, and does not specify how to make it (based on what input).

So Scripting needs to define how does it handle input for creating exposed Things. Currently it is a dictionary object that undergoes the input validation algorithm, that maps to this "partial TD" term.

Since Scripting is optional and is just a client, we (TD TF, Arch TF) still might need a broader definition on how to remotely create an exposed Thing, what input is accepted and how. Therefore the Arch TF could consider if this term is needed, despite being used only in Scripting (and actually not even there, as we separated the terminology/TD/partial TD validation problem from the JavaScript input type/validation problem).

mlagally:
I think we should preserve the definition in the terminology section of the architecture spec.
This may come at hand in other areas, e.g. discovery, thing directories, TD, ...

Issue #595, opened 2021-06-29
Chapter 7.1.2 - rework "Thing Model" to new title/content  "Instance descriptions and class models"


mlagally:
Chapter 7.1. should contain a section called "Instance descriptions and class models"

This is an introduction to the concepts, i.e. descriptive documents to describe a class of things and instances.
Explain that we have a modeling language and a description for individual instances.
Devices can be self descriptive, i.e. carry their TD.

PR #600, opened 2021-06-29
Add Description and Model Section

mmccool:
Resolves https://github.com/w3c/wot-architecture/issues/595

- [x] rename section as discussed
- [x] assign more relevant id
- [x] cross-reference issue
- [x] add content
- [ ] resolve whether (and if so, where) we should describe using TMs for incomplete descriptions

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/600.html" title="Last updated on Sep 23, 2021, 2:34 PM UTC (c855f01)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/600/b5dba58...mmccool:c855f01.html" title="Last updated on Sep 23, 2021, 2:34 PM UTC (c855f01)">Diff</a>


mmccool:
I suspect I should hold off before working on this until I see what @sebastiankb added around TMs.

mmccool:
- Added link to issue and a discussion paragraph
- I'm not entirely happy with the content since the section should introduce the *requirement* for instance descriptions and class descriptions, but also immediately proposes that TMs can be used to describe instances incompletely
- We also jump directly into TDs and TMs directly when I think this section should be talking about general requirements
- I think a cleaner structure would talk about instances, classes, and the occasional need for "incomplete instance descriptions"; then LATER in the text we should introduce TDs (and say they can be used for complete instance descriptions) and TMs (and say they can be used both for classes and incompete instance descriptions).  
- Although, really, using TMs for incomplete instance descriptions is a hack IMO.  It arises from the fact that "forms" are mandatory.  However, we discussed this just yesterday in the TD call for MQTT and some other standards (e.g. ECHONET) where the communication metadata is implicit and agreed to make forms optional.  So maybe we can just drop this particular usage of TMs.
- The usage when TMs appear before an instance is fully configured is however consistent with their use for describing a class.

mmccool:
Anyway, still marking as a WIP until we sort out the above.

mmccool:
I was also thinking we could rename this section "Metadata" to be more exclusive, and that might avoid the conceptual conflict with using TMs for incomplete instance descriptions, too.

mmccool:
Discussion:
- Ben; virtual devices vs. services.  If have a hammer, everything is a nail.  Don't necessarily want to remodel the web to fit into WoT model.
- Text currently does NOT talk about TDs for services
- MM: most appropriate for services associated with a device, eg. edge AI service; handy for mashups if everything is using same description format
- MK: TD is good for "smart object" pattern, beyond IoT; can't restrict people anyway, but don't have to add changes
- MM: edge is compute resources *at a particular location*
- ML: support including intelligent edge use cases

Issue #598, opened 2021-06-29
Section 8.3 - device categories


mlagally:
link should be removed 
Class 4 should include gateways / hubs

Issue #581, opened 2021-02-25
Consolidate usage of gateway, edge and hub


mlagally:
Arch call on 25.2.:
The current arch. document sometimes uses the term "gateway", "edge" in an inconsistent way.
We need to consolidate the text after we have defined the term "hub".

"hub" could be the generic term.

New terminology:
hub: A hub is a hardware platform that centralizes multiple services, such as proxies, firewalls, digital twins, shadows, intermediaries, orchestration, directories, dashboards, and other services.

shadow: reflects the state of another device.

we should also consider aligning with the term "intermediary".

After that the architecture diagram needs probably be updated.

PR #601, opened 2021-06-29
Fixup Device Classes

mmccool:
Resolves https://github.com/w3c/wot-architecture/issues/598

- [x] fix reference to RFC7228
- [x] use "gateway/hub" for Class 4
- [x] Move section to more appropriate location (it's not really a building block; perhaps move under Terminology?)
- [x] General grammar and positioning cleanup (e.g. sentence in which RFC is cited is confusing)

Note: "hub" is only mentioned in this section now and no-where else.  So the definition of hub also needs to be addressed, but this should be dealt with in a different PR that addresses issue https://github.com/w3c/wot-architecture/issues/581

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/601.html" title="Last updated on Jul 15, 2021, 1:46 PM UTC (a5dfece)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/601/af1a8df...mmccool:a5dfece.html" title="Last updated on Jul 15, 2021, 1:46 PM UTC (a5dfece)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -731,6 +731,80 @@
 
   </section>
 
+  <!-- Device categories -->
+  <section id="device-categoriesdevices">
+      <h2>Device Categories</h2>	  
+      <p>
+        In a deployment of WoT conforming to the <a href="https://www.w3.org/TR/wot-architecture/#architecture-abstract">WoT Abstract Architecture</a>
+        we see different device types with typical characteristics.
+        They range (sorted in the order of footprint and capabilities) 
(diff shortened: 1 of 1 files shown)


mmccool:
When revising the text I found I was not entirely happy with the device categories, so added some comments for some other possible names and also an ednote discussing some other options.  We could either merge this as is, or discuss (and possibly revise) the names and add a separate section on security (rather than including them in the names) and then remove the ednote.

Issue #585, opened 2021-03-22
Architectural patterns: Introduce star(hub) vs. p2p section


mlagally:
The ITU-T specifications introduce a broker concept, which is implicitly also enabled in W3C WoT.
A new section should explain this and make it more explicit.

This could go into section 7 or section 10.

Please discuss pro's / cons in this issue.

PR #603, opened 2021-06-29
Define and Discuss "Hubs"

mmccool:
* Resolves https://github.com/w3c/wot-architecture/issues/581
* Resolves https://github.com/w3c/wot-architecture/issues/585

- [x] add "hubs" to list of Edge Devices
- [x] expand comment on what to do for Edge Device discussion
- [ ] update Edge Device section to discuss hubs vs. gateways
- [ ] add explicit definitions for hubs and gateways to terminology
- [ ] make sure use of "Gateway" in "Smart Home Gateway" is consistent with definition of "Gateway" (e.g. if it manages local devices, and is NOT just a proxy... maybe it should be a "Smart Home Hub"?
- [ ] Explicitly allow for star as well as p2p topologies in architectural patterns (addressing https://github.com/w3c/wot-architecture/issues/585)
- [ ] cite use of hubs as a place to locate "digital twins", and along the way clarify the various functions that a digital twin might perform, and which are optional (e.g. protocol conversion, state caching, simulation, prediction, etc).  A "simple" digital twin might just do protocol conversion and state caching, for instance, and these simple twins are commonly supported by hubs.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/603.html" title="Last updated on Jul 15, 2021, 4:18 PM UTC (432a5cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/603/af1a8df...mmccool:432a5cd.html" title="Last updated on Jul 15, 2021, 4:18 PM UTC (432a5cd)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/603.html" title="Last updated on Mar 24, 2022, 10:43 AM UTC (432a5cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/603/af1a8df...mmccool:432a5cd.html" title="Last updated on Mar 24, 2022, 10:43 AM UTC (432a5cd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -452,7 +452,7 @@
       </dt>
       <dd>A device that provides an entry point into
         enterprise or service provider core networks. Examples
-        include gateways, routers, switches, multiplexers, and a
+        include hubs, gateways, routers, switches, multiplexers, and a
         variety of other access devices.</dd>
       <dt>
         <dfn>Event</dfn>
@@ -1131,6 +1131,7 @@
       <section class="ednote" title="TODO">
 	<h3>TODO</h3>
         This section will be expanded to capture recent activities.
+        In particular, it would be a good place to discuss hubs and "star" topologies.
         <!-- McCool -->
       </section>
 


mmccool:
So I realized a more extensive refactoring is needed here, and I wanted to discuss the plan before proceeding.  In particular, we have a section called "Smart Home Gateways" that may need to be renamed, and it's not clear if it's talking about hubs or *just* gateways.  Also, Gateways are not limited the smart home.  The following section talks about "edge devices" and the definition implies this includes hubs and gateways, but elsewhere in the text we use this term only for "larger" edge computers (e.g. that provide advanced services).  So one option here would be to add a "hub" section in between "Gateways" and "Edge Devices" (renaming "Smart Home Gateways" to just "Gateways"), then focus each section on one particular level (e.g. refactoring hub functionality out of the current gateway section into the new hub section...)

mryuichi:
First of all, it's a good idea to rename the session title "smart home gateway" to "gateway".  In the first draft, all the descriptions of the topology used the smart home as an example.  Only the name at that time remained.

I'm confused because of the difference between hub and gateway. The word "hub" is rarely used in the current documents.
We should define gateway, hub, and edge. In terminology session, hub, gateway, switch, and router are exemplified as the definition of Edge devices. These are clearly different from a network perspective but similar from a computer perspective. The definition of the edge device is to provide an entry point. Then it should be defined from the perspective of the computer that provides the entry point, and should not include anything transparent to computers like a switch or router. As a result, the edge device must be a device and a gateway that has an entry point.

The device is represented by a TD and has an entry point. On the other hand, the gateway provides an entry point instead of the device to be connected. There are two cases where the gateway provides it. One is that the adapter on the gateway provides the TD and entry point instead of the device that has a legacy interface. The other is that a shadow on the gateway provides the TD and entry point. This enhances security by avoiding direct access to the device from the consumers. Fujitsu's gateway supports both of them and the former is called "gateway" and the latter is "proxy". Is "hub" one of these?

mmccool:
Related to some work in use cases on hubs, e.g. description of EdgeX.  This is an outlier, and really use cases should look at a survey of "hub" architectures.  At any rate, it would be good to coordinate terminology, etc.  And yes, the term "hub" is not ideal, but I think it's better than "gateway", at least in the smart home use case.

See: https://github.com/w3c/wot-usecases/pull/158

mlagally:
Arch call on March 3rd:
Suggest to decouple the remaining tasks and merge this MR.
Check with @mmccool

PR #624, opened 2021-11-02
preparing implementation report

mlagally:
Extracted list of RFC2119 assertions

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/pull/624.html" title="Last updated on Nov 4, 2021, 10:29 AM UTC (49bee48)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/624/6531e45...49bee48.html" title="Last updated on Nov 4, 2021, 10:29 AM UTC (49bee48)">Diff</a>


egekorkan (changes requested):
I do not think that there is a need to insert Not testable by Assertion Tester. It is based on TDs so it is not applicable here.

egekorkan:
Other points: 

- [ ] `arch-hypermedia-controles` has no text attached. Weird behavior. This assertion is found in the spec document but it is inside a header?
- [ ] Line 45 has an unidentified id
- [ ] Lines 44-47 have assertion ids that do not start with `arch`
- [ ] `discovery-metadata-after-authorization` has no text associated

k-toumura:
> I do not think that there is a need to insert Not testable by Assertion Tester. It is based on TDs so it is not applicable here.

Yes.  Instead of putting "Not Testable ..." in the third column, it might be better to simply put an assertion statement in the third column as a comment.
By the way, does the report generation tool safely ignore the third (and fourth) column of manual.csv?

mlagally:
> > I do not think that there is a need to insert Not testable by Assertion Tester. It is based on TDs so it is not applicable here.
> 
> Yes. Instead of putting "Not Testable ..." in the third column, it might be better to simply put an assertion statement in the third column as a comment. By the way, does the report generation tool safely ignore the third (and fourth) column of manual.csv?
> 
> ```
> "ID","Status","Comment"
> "arch-td-metadata","null","Assertion: In W3C WoT, the description metadata for a Thing instance MUST be available as a WoT Thing Description (TD) [[?WOT-THING-DESCRIPTION]]."
> ```

I think we can improve the csv format generation later - for now the columns ID and assertion are relevant.

egekorkan (approved):

egekorkan:
Also I think that requesting review from 5 people is not a good idea. It is never clear if all the 5 people should make a review (an AND relation) or 1 is enough (an OR relation). When it is an OR relation, then no one knows if they have the responsibility or not, reducing chances for getting a review

k-toumura (approved):

PR #669, opened 2021-12-23
Producer definition

sebastiankb:
Producer definition for terminology section

### DIFF
--- a/index.html
+++ b/index.html
@@ -590,6 +590,12 @@
         and is only used via abstract operations, such as signing, that do not reveal the secret
         information even to the application using it.</dd>
       <dt>
+        <dfn>Producer</dfn>
+      </dt>
+      <dd>An entity that can create WoT Thing Descriptions
+        for a specific Thing.</dd>
+      <dt>
+      <dt>
         <dfn>Property</dfn>
       </dt>
       <dd>An Interaction Affordance that exposes state of the Thing.


mlagally:
@sebastiankb 
"... for a specific Thing" sounds like a limitation and I think it is more generic.

Suggest "for a specific Thing or multiple Things" or to just remove that part of the definition.

index.html
```diff
@@ -589,6 +589,12 @@ <h1>Terminology</h1>
         system. Ideally such data is stored in a separate memory inaccessible to the application
         and is only used via abstract operations, such as signing, that do not reveal the secret
         information even to the application using it.</dd>
+      <dt>
+        <dfn>Producer</dfn>
+      </dt>
+      <dd>An entity that can create WoT Thing Descriptions
+        for a specific Thing.</dd>
```
mlagally, line 596: Is this a limitation?

mlagally (changes requested):

egekorkan:
> "... for a specific Thing" sounds like a limitation and I think it is more generic.

Well, a Producer can write multiple TDs but a TD is for a single Thing. But there is an extra s there I think.

mlagally:
Arch call on Jan 27th:

Other specifications use "Exposer", relationship should be defined. Please seek alignment.
Exposer does not imply generating a TD.
"TD Producer" could be used to make it more unambiguous
A directory service is a producer, however the current text talks about "a thing" which could be read as a limitation.

index.html
mmccool: This exactly corresponds to the term "Generator" that I proposed in my comment.  I'm ok with "Producer" under this specific definition, however.  As long as we are clear that "Exposer != Provider != Producer", since these might all be separate entities.

benfrancis:
This definition looks OK to me, but how about simply "An entity which generates WoT Thing Descriptions."

It would simpler if there were just two roles:
- Producer - Generates and exposes the metadata and interfaces of a Thing
- Consumer - Processes and interacts with the metadata and interfaces of a Thing

The reality seems to be more complex, and the roles can be broken down as (paraphrasing):
- Thing - An abstraction of a physical device whose capabilities and interfaces are described by a Thing Description
- Thing Description - Describes the capabilities and interfaces of a Thing
- Producer - Produces Thing Descriptions for one or more Things
- Exposer - Exposes Thing Descriptions for one or more Things
- Consumer - Processes Thing Descriptions and interacts with Things

PR #696, opened 2022-02-01
New section on "virtual things", clarifying terminology

### LINKED ISSUE #682: clarify virtual thing
Arch call on 27.1.:
A virtual thing can be a room or a basic simulation.

Need to clarify terminology, the current terminology definition does not cover several use cases.

Seek alignment with publicly known definitions
https://en.wikipedia.org/wiki/Virtual_device Virtual device
https://en.wikipedia.org/wiki/Digital_twin Digital twin

mlagally:
fixes #682

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/pull/696.html" title="Last updated on Feb 1, 2022, 3:16 PM UTC (df70bbb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/696/7368123...df70bbb.html" title="Last updated on Feb 1, 2022, 3:16 PM UTC (df70bbb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -669,8 +669,7 @@
       </dt>
       <dd>An abstraction of a physical or a virtual entity
         whose metadata and interfaces are described by a WoT
-        Thing Description, whereas a virtual entity is the
-        composition of one or more Things.</dd>
+        Thing Description.</dd>
       <dt>
         <dfn data-lt="WoT Thing Description Directory">Thing Description Directory</dfn> 
       </dt>
@@ -714,8 +713,11 @@
       <dt>
         <dfn>Virtual Thing</dfn>
       </dt>
-      <dd>An instance of a Thing that represents a Thing that is located
-        on another system component.</dd>
+      <dd>A virtual entity that represents a thing that conforms to a TD 
+          but does not have a direct physical representant. It may be a software simulation,
+	      a composition of one or more Things which may be located on different system components, 
+	      or an abstract entity, such as a named location or a geofence.	    
+	  </dd>
       <dt>
         <dfn>Vocabulary</dfn>
       </dt>
@@ -1420,16 +1422,20 @@
     </section>
 
     <section id="virtual-thing">
-      <h3>Virtual Things</h3>    
-      <section class="ednote" title="TODO">
-      <h3>TODO</h3>
-      Create a description and a diagram for a <a>virtual thing</a> scenario,
-      as described in the use case for 
-      <a href="https://w3c.github.io/wot-usecases/#USE-CASES/Retail-virtual-thing">Virtual Thing</a>
+      <h3>Virtual Things</h3>   
+	<p>
+	 A virtual thing is an abstraction of a physical entity that does not have a direct physical counterpart.
+	 It can be used to describe abstract interfaces with a TD that are decoupled from individual devices. 
+	 The purpose is to decouple a functional interface that defines the behavior
+	 from the specific implementations of devices among several manufacturers.
+	 The interface remains stable and is independent from the actual implementation of things behind that interface.
+	</p> 
+	<p>
+	  The virtual thing guarantees that 
+	 devices can be replaced without affecting the implementation of the consumer, 
+	 who can continue using the device as he did before.
+	</p>    
       </section>
-    </section>
-
-
     <section>
       <h3>Cross-domain Collaboration</h3>
       <p>


index.html
```diff
@@ -1420,16 +1422,20 @@ <h3>Multi-Cloud</h3>
...
-      <h3>Virtual Things</h3>    
-      <section class="ednote" title="TODO">
-      <h3>TODO</h3>
-      Create a description and a diagram for a <a>virtual thing</a> scenario,
-      as described in the use case for 
-      <a href="https://w3c.github.io/wot-usecases/#USE-CASES/Retail-virtual-thing">Virtual Thing</a>
+      <h3>Virtual Things</h3>   
+	<p>
+	 A virtual thing is an abstraction of a physical entity that does not have a direct physical counterpart.
+	 It can be used to describe abstract interfaces with a TD that are decoupled from individual devices.
```
egekorkan, line 1428: Until this line I was OK but then after this line, it sounds too much like TM. It is also a bit wrong to mix the two.

index.html
```diff
@@ -1420,16 +1422,20 @@ <h3>Multi-Cloud</h3>
...
-      <section class="ednote" title="TODO">
-      <h3>TODO</h3>
-      Create a description and a diagram for a <a>virtual thing</a> scenario,
-      as described in the use case for 
-      <a href="https://w3c.github.io/wot-usecases/#USE-CASES/Retail-virtual-thing">Virtual Thing</a>
+      <h3>Virtual Things</h3>   
+	<p>
+	 A virtual thing is an abstraction of a physical entity that does not have a direct physical counterpart.
+	 It can be used to describe abstract interfaces with a TD that are decoupled from individual devices. 
+	 The purpose is to decouple a functional interface that defines the behavior
```
egekorkan, line 1429: interface is not defined and can be interpreted differently by different people

index.html
```diff
@@ -1420,16 +1422,20 @@ <h3>Multi-Cloud</h3>
...
+	 from the specific implementations of devices among several manufacturers.
+	 The interface remains stable and is independent from the actual implementation of things behind that interface.
+	</p> 
+	<p>
+	  The virtual thing guarantees that 
+	 devices can be replaced without affecting the implementation of the consumer, 
+	 who can continue using the device as he did before.
```
egekorkan, line 1436: A better wording around the gender is needed

index.html
```diff
@@ -1420,16 +1422,20 @@ <h3>Multi-Cloud</h3>
...
+      <h3>Virtual Things</h3>   
+	<p>
+	 A virtual thing is an abstraction of a physical entity that does not have a direct physical counterpart.
+	 It can be used to describe abstract interfaces with a TD that are decoupled from individual devices. 
+	 The purpose is to decouple a functional interface that defines the behavior
+	 from the specific implementations of devices among several manufacturers.
+	 The interface remains stable and is independent from the actual implementation of things behind that interface.
+	</p> 
+	<p>
+	  The virtual thing guarantees that
```
egekorkan, line 1434: Who gives this guarantee? It sounds a little too "serious" for a document/chapter that is this "loose".

egekorkan (changes requested):

index.html
mlagally: @egekorkan 
Please be more specific about your concerns. Please review the virtual thing use case in the use case document.

index.html
mlagally: I don't understand your concern. Please propose text.

index.html
mlagally: Please elaborate what you mean with "too serious"

index.html
```diff
@@ -1420,16 +1422,20 @@ <h3>Multi-Cloud</h3>
...
     <section id="virtual-thing">
-      <h3>Virtual Things</h3>    
-      <section class="ednote" title="TODO">
-      <h3>TODO</h3>
-      Create a description and a diagram for a <a>virtual thing</a> scenario,
-      as described in the use case for 
-      <a href="https://w3c.github.io/wot-usecases/#USE-CASES/Retail-virtual-thing">Virtual Thing</a>
+      <h3>Virtual Things</h3>   
+	<p>
+	 A virtual thing is an abstraction of a physical entity that does not have a direct physical counterpart.
```
mmccool, line 1427: could just say "a virtual thing is an abstraction that does not have a direct physical counterpart".  I had to scratch my head over the "abstraction of a physical entity that does not have a direct physical counterpart".  So does it have a physical referent, or not?  I think you're trying to say it does not, so...

index.html
mmccool: I agree, "abstract interfaces" sounds too much like what a TM is trying to do.  It may be explained elsewhere, but that does not mean this statement isn't confusing.  If you remove the word "abstract" it might be ok.  For example, it could be a software-only service or simulation, which has an interface but is not a device.  To me "abstract" means "without an instantiation" which describes a Thing Model, not a running simulation or service (which is instantiated.  It is virtual, but not abstract).

index.html
mmccool: I think the issue here is that the spec is making a guarantee but is not in a position to enforce it.  We could say that *one* use case is to provide a stable interface as a wrapper to another service that may be subject to change.  It's an application of virtual things, not its sole purpose or something we can guarantee.  I could also have a virtual thing that is a software service that is changing all the time.  It may not be a good idea, but I might want to do it for various reasons.

index.html
egekorkan: > @egekorkan Please be more specific about your concerns. Please review the virtual thing use case in the use case document.

I had a look and the main issue seems to be the use cases document. Should I open an issue there?

index.html
mmccool: We should avoid "he" since it projects a gender on the user for no good reason.  Just replace with "they", which works fine here.  I need to look at the context, but if "consumer" is the Consuming Thing, then "it" may also be appropriate.

egekorkan: replacing he with a gender-neutral pronoun/subject is needed or the subject should be removed. proposal:

`who can continue using the device as before.`

mmccool: Actually, if you mean the *person* using the thing, using the word "user" would avoid confusion with Consumer.  If you mean Consumer, capitalize and use "it".

index.html
egekorkan: Well this section is non normative so we should not talk about any guarantees

egekorkan:
I think the main issue is that this PR contains the text of the use cases document for a use case that is submitted by Connexxus. The wording in the use cases document is confusing (mixing virtual things, TMs and composite Things/TD) into a single definition without realizing. We do not have to copy paste from use cases and instead write it in a way that fits the other concepts we have in the WoT

mlagally:
Arch call on 10.3.:
Since we had a later update on terminology, closing this MR without merging.

PR #701, opened 2022-02-03
Update Implementation Report link in SOTD

### LINKED ISSUE #699: Update discussion of Implementation Reports in SOTD
The "Status of this Document" links to just the TD implementation report but we will have multiple deliverables with multiple IRs in this round, and (probably) also one for Architecture.  This needs to be updated, I suggest to point at just the *Architecture* IR and in that report we can point at the others if we want.  Although we could also link to all of them in this section, the purpose of this section is to talk about the status of the *Architecture* document, so...

mmccool:
Resolves #699 
- Links to (new) Architecture IR (still to be created...) in SOTD
- Links to Architecture IR in header, consistent with other docs
- Says that Architecture IR describes implementations following the architecture
- Says that Architecture IR links to other IRs for building blocks
So when we do write the Arch IR, we need to remember to do the last two points.  The alternative would be to link to each of the building blocks' IRs directly from the Architecture SOTD but I think that will just clutter things up.  They are also linked from the header of each document (or should be).

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/701.html" title="Last updated on Feb 3, 2022, 2:31 AM UTC (4f83abb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/701/7368123...mmccool:4f83abb.html" title="Last updated on Feb 3, 2022, 2:31 AM UTC (4f83abb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -18,6 +18,7 @@
       shortName: "wot-architecture11",
       copyrightStart: 2017,
       wgPublicList: "public-wot-wg",
+      implementationReportURI: "https://w3c.github.io/wot-architecture/testing/report.html",
       github: {
         repoURL: "https://github.com/w3c/wot-architecture",
         branch: "master"
--- a/index.html
+++ b/index.html
@@ -285,10 +285,11 @@
     <p>
       This document describes an abstract architecture design.
       However, there is an
-      <a href="https://w3c.github.io/wot-thing-description/testing/report.html">Implementation Report</a>
-      that describes a set of concrete implementations based on the associated <em>WoT Thing Description</em>
-      specification.
-      These are implementations following the W3C Web of Things architecture.
+      <a href="https://w3c.github.io/wot-architecture/testing/report.html">Implementation Report</a>
+      that describes a set of concrete implementations 
+      following the W3C Web of Things architecture.
+      It also references the other implementation
+      reports for the various WoT building blocks.
     </p>
   </section>
   <section id="introduction" class="informative">


mlagally:
@mmccool 
I don't think we should remove the link to the implementation report of the TD. 
In the first release we argued that an implementation report for an abstract architecture is not required and we point to the implementation report of the TD. 
Now we will also create an implementation report for the profile, which is the second REC track specification that will be validated in a test fest.
I think this assumption still holds that a dedicated implementation report for an abstract architecture is not required and we should reference the implementation reports for the TD and the profile. 
We should follow the KISS principle and reference these reports directly from the architecture document.

mmccool:
Observations:
1. We will have three normative deliverables now with IRs (Profile, Discovery, TD)
2. The wide review process expects an IR.  Last time we could point to just one other one but it will complicate the wide review process to point to three (and increase the workload of reviewers, which will slow things down).
3. Implementations of each deliverable != implementations of the WoT architecture as a whole.  Is an implementation of a WoT Directory service the same as the implementation of the WoT Architecture?  Is a single Thing?  I think it would be useful to have a place to describe systems that instantiate the WoT Architecture as a whole (it's an abstract architecture, but we can describe concrete instantiations such as Desigo CC, Takenaka's BIM system, etc.  It's not necessary that all systems implement all aspects of the Architecture, however...).
4. We do have assertions in the Architecture spec still and need a place to report that there are at least two systems in existence that satisfy each one.

Because of 4 in particular I think we DO need a separate Architecture IR.  It would also make 2 easier (the alternative is to ask reviewers to read all three, and THEN they would find out that the assertions in the Arch doc are not covered in any of them...), and provides a central place to put 3.

So here are some possible solutions (however, note I feel 0 is not feasible for the reasons noted just above):
0*. Do not have an Arch IR and just cite each of the deliverable IRs.
1. Cite each of the deliverable IRs in the Arch document as well as citing a (simple) Arch IR.
2. Instead of 1, cite the deliverable IRs in the Arch IR and cite the Arch IR in the Arch spec (what this PR is assuming).
3. Do both 1 and 2.  This will make it easier for reviewers to find the various IRs (although since they are also in the headers of each document this is not that hard).

I am personally OK with any of 1, 2, or 3.

One additional issue though is that the IRs are not really "published" documents in the same way as Notes, etc.  Currently we are citing them as direct links into github.io, which frankly is not ideal.  We should consider whether these should be citations/references, how to deal with versions (e.g. report11.html or report.html), etc.  If the links change for whatever reason it would be helpful to minimize the number of places they need to be updated.  For instance we could cite just the docs themselves since the IRs are linked from the headers (and we have proper references for these documents).

Issue #709, opened 2022-02-15
Definitions for Virtual Thing, Shadow, and Digital Twins


mmccool:
Proposed definition for Virtual Thing: A kind of Service that represents another Thing.

Examples are Digital Twins and Shadows.
* Shadows keep a copy of state of other Things. 
* Digital Twins may do that too (keep a copy of state of other Things) but may also have the ability to predict future states or behaviour, and may also maintain a history of past state (suggest a list of additional functionalities for Twins that if an entity has at least one can be called a Twin).

Note: the use of Virtual Thing for a "composite of other Things" should be deprecated and we should introduce a new term for that, perhaps "Composite Thing"

mmccool:
Two additional thoughts:
- Proxies are another example of a Virtual Thing.  Proxies are Services that re-expose the same network interface as another Thing, but with modified access details (URLs, security schemes, etc).
- Regarding "Composite Things", I can imagine Composite Things that are ALSO Virtual Things.  I think a "Composite Thing" is an entity that has a Thing Description and describes the *relationship* between a set of Things.  If there is also a related Service that has a network interface that allows interactions with that set (for example, a Composite Thing might represent the set of all the lights in a room, and ALSO there might be a Service that allows all the lights to be turned on and off at once), the Composite Thing is also a Virtual Thing.  But I think the definition of the Composite Thing should be specifically about "An entity representing a set of relationships between Things", i.e. the *concept* of a collection, without implying there is a Service or Device with a network interface related to that collection (there may not be).

Propose we discuss whether these two definitions belong in this PR/issue or a new one.  For now I'll just do the ones in the description.

mmccool:
@relu91 to avoid confusion with other interpretations of Virtual Thing, I am proposing the specific term "Service" to represent a Thing that is associated with a software entity instead of a physical device.  An example would be a TDD.  The fact is we already use "service" a lot for this.

A Virtual Thing is then a special case of a Service that "represents another Thing", examples being Shadows and Twins (and maybe Proxies).

We have also used "Virtual Thing" to represent a *conceptual* collection that may not have a network interface, or a representation of another entity like a location.  I think these are valid cases of "Things" but would like to exclude the use of "Virtual Thing" for these unless there is an associated Service.  We can invent new names for these kinds of Thing (e.g. Composite Thing, Location Thing) as needed.

Also, everything with "Thing" in the name is a Thing.  I think a Venn diagram would be useful...

mmccool:
[...]

Issue #708, opened 2022-02-15
Definitions for Entities that are Things


mmccool:
Add two definitions:
- Device is a physical entity that has a network interface.  Devices can be described by a Thing Description and are a kind of Thing.
- Service is a software entity that has a network interface.  Services can be described by a Thing Description and are a kind of Thing.

The intention is that both Devices and Services, since they can be described by Thing Descriptions, are Things.

mmccool:
Can also propose a definition for Virtual Thing: a kind of service that represents another Thing.  

Examples are Digital Twins and Shadow.   Shadows keep a copy of state of other Things.  Digital Twins may do that but may also have the ability to predict future states or behaviour, and may also maintain a history of past state (suggest a list of additional functionalities for Twins that if an entity has at least one can be called a Twin).

Will not include these in this PR, just putting out there for discussion.

ashimura:
As I mentioned during the WoT Editors call yesterday on Feb. 15, it might make sense to look into the definitions and acronyms by oneM2M again:
* https://onem2m.org/using-onem2m/developers/basics
* https://onem2m.org/images/files/deliverables/TS-0011-Common_Terminology-V1_2_1.pdf

k-toumura:
As a starting point for discussion, I've drawn a diagram of the Thing-related terms being discussed.
![terms](https://user-images.githubusercontent.com/30313213/154259573-b8bdc5ed-45ea-403c-b17f-09908694221f.svg)

In this diagram, I've separated the Physical Device from the software that makes them network-accessible (Physical Things).

(The diagram is attached as SVG file, so please feel free to fetch, rewrite, and upload it for discussion.)

mmccool:
Thanks for the diagram.  Let's discuss next time we get a chance... but I'm thinking something even simpler, like a UML diagram showing subclass relationships among our definitions, would be helpful.

mlagally:
This is a great diagram and unfortunately I just see it now after our architecture call. 
We should include it in the architecture specification, need to find a good plave.

PR #711, opened 2022-02-16
Definitions for Device, Service, etc.

mmccool:
New definitions for Device, Service, and Shadow.
Updated definitions for Virtual Thing and Digital Twin.
Resolves issue #709.
Resolves issue #708.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/711.html" title="Last updated on Feb 17, 2022, 5:15 PM UTC (14d0d30)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/711/78521c3...mmccool:14d0d30.html" title="Last updated on Feb 17, 2022, 5:15 PM UTC (14d0d30)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -447,17 +447,16 @@
       <dt>
         <dfn>Digital Twin</dfn>
       </dt>
-      <dd>A digital twin is type of <a>Virtual Thing</a>,
-        specifically
-        a <a>Service</a> that resides on a cloud or edge node 
-	and represents another device or a group of devices.
-	Digital Twins can be used to represent and provide 
+      <dd>A digital twin is type of <a>Virtual Thing</a>
+        that resides on a cloud or edge node. 
+	Digital Twins may be used to represent and provide 
 	a network interface for
-        real-world devices which may not be continuously online,
-        or to run simulations of new applications and services
-        before they get deployed to the real devices.
-	They may also be used to maintain a history of past
-	state or behviour or to predict future state of behaviour.
+        real-world devices which may not be continuously online
+	(see also <a>Shadows</a>),
+        may be able to run simulations of new applications and services
+        before they get deployed to the real devices,
+	may be able to maintain a history of past state or behaviour, 
+	and may be able to predict future state or behaviour.
 	Digital Twins typically have more functionality than
 	simple <a>Shadows</a>.
       </dd>
@@ -646,6 +645,18 @@
         blocks. A Servient can host and expose Things and/or host Consumers that consume Things.
         Servients can support multiple Protocol Bindings to enable
         interaction with different IoT platforms.</dd>
+      <dt>
+        <dfn>Shadow</dfn>
+      </dt>
+      <dd>A Shadow is a <a>Virtual Thing</a> that
+      maintains a copy of the state and mediates interactions with another
+      <a>Thing</a> which may not be immediately available due to network
+      connectivity, power savings modes, or other factors.
+      A Shadow aims to achive eventual consistency with the state
+      of the Thing it represents.
+      If a Shadow has more advanced functionality than simply caching state
+      it may be better to refer to it as a <a>Digital Twin</a>.
+      </dd>
       <dt>
         <dfn>Subprotocol</dfn>
       </dt>
--- a/index.html
+++ b/index.html
@@ -4319,13 +4319,15 @@
 
     <h2 id="changes-from-recommendation-1.0">Changes from the 1.0 version of [[wot-architecture]]</h2>
     <ul>
-      <li>Add definitions for Device, Service, and Shadow.  Update definitions for
-	  Virtual Thing and Digital Twin.  Narrow definition of Virtual Thing to
-	  a Service that represents another Thing.</li>
+      <li>Add definitions for Connected Device (a.k.a. Device), Service, and Shadow.  
+	  Update definitions for
+	  Virtual Thing and Digital Twin.  
+	  Narrow definition of Virtual Thing to
+	  a Service that mediates with one of more other Things.</li>
       <li>Split Security and Privacy Considerations into separate chapters.</li>
       <li>New chapter: Lifecycle.</li>
       <li>New terminology: Thing Model.</li>
-      <li>Chapter restructuring and renaming
+      <li>Chapter restructuring and renaming:
         <ul>
           <li>Application Domains (Verticals)</li>
           <li>System Topologies (Horizontals)</li>
--- a/index.html
+++ b/index.html
@@ -4304,10 +4304,13 @@
 
     <h2 id="changes-from-recommendation-1.0">Changes from the 1.0 version of [[wot-architecture]]</h2>
     <ul>
+      <li>Add definitions for Device, Service, and Shadow.  Update definitions for
+	  Virtual Thing and Digital Twin.  Narrow definition of Virtual Thing to
+	  a Service that represents another Thing.</li>
       <li>Split Security and Privacy Considerations into separate chapters.</li>
       <li>New chapter: Lifecycle.</li>
       <li>New terminology: Thing Model.</li>
-      <li>Chapter restructuring and renaming:
+      <li>Chapter restructuring and renaming
         <ul>
           <li>Application Domains (Verticals)</li>
           <li>System Topologies (Horizontals)</li>
--- a/index.html
+++ b/index.html
@@ -445,6 +445,12 @@
         and corresponding data items that are passed between <a>Things</a>
         and <a>Consumers</a> during interactions.</dd>
       <dt>
+        <dfn>Device</dfn>
+      </dt>
+      <dd>A Device is a physical entity that has a network interface.
+      Devices can be described by a <a>Thing Description</a> and are a kind of <a>Thing</a>.
+      Compare with <a>Service</a>.
+      <dt>
         <dfn>Digital Twin</dfn>
       </dt>
       <dd>A digital twin is type of <a>Virtual Thing</a>
--- a/index.html
+++ b/index.html
@@ -669,11 +669,14 @@
       </dt>
       <dd>A Shadow is a <a>Virtual Thing</a> that
       maintains a copy of the state and mediates interactions with another
-      <a>Thing</a> which may not be immediately available due to network
+      <a>Thing</a>.
+      <!--
+      which may not be immediately available due to network
       connectivity, power savings modes, or other factors.
-      A Shadow aims to achive eventual consistency with the state
+      -->
+      A Shadow aims to achieve eventual consistency with the state
       of the Thing it represents.
-      If a Shadow has more advanced functionality than simply caching state
+      If a Shadow has more functionality than simply mirroring state
       it may be better to refer to it as a <a>Digital Twin</a>.
       </dd>
       <dt>
--- a/index.html
+++ b/index.html
@@ -447,12 +447,20 @@
       <dt>
         <dfn>Digital Twin</dfn>
       </dt>
-      <dd>A digital twin is a virtual representation of a
-        device or a group of devices that resides on a cloud
-        or edge node. It can be used to represent
+      <dd>A digital twin is type of <a>Virtual Thing</a>,
+        specifically
+        a <a>Service</a> that resides on a cloud or edge node 
+	and represents another device or a group of devices.
+	Digital Twins can be used to represent and provide 
+	a network interface for
         real-world devices which may not be continuously online,
-        or to run simulations of new applications and services,
-        before they get deployed to the real devices.</dd>
+        or to run simulations of new applications and services
+        before they get deployed to the real devices.
+	They may also be used to maintain a history of past
+	state or behviour or to predict future state of behaviour.
+	Digital Twins typically have more functionality than
+	simple <a>Shadows</a>.
+      </dd>
       <dt><dfn data-lt="WoT Discovery">Discovery</dfn>
       <dd>Mechanisms defined by WoT for distributing and accessing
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -716,8 +748,10 @@ <h1>Terminology</h1>
       <dt>
         <dfn>Virtual Thing</dfn>
       </dt>
-      <dd>An instance of a Thing that represents a Thing that is located
-        on another system component.</dd>
+      <dd>A <a>Service</a> that represents another <a>Thing</a>.
+       A Virtual Thing will often as as an <a>Intermediary</a>.
```
benfrancis, line 752: typo: "as as"

index.html
```diff
@@ -444,15 +444,28 @@ <h1>Terminology</h1>
       <dd>A data schema describes the information model and the related payload structure
         and corresponding data items that are passed between <a>Things</a>
         and <a>Consumers</a> during interactions.</dd>
+      <dt>
+        <dfn>Device</dfn>
+      </dt>
+      <dd>A Device is a physical entity that has a network interface.
+      Devices can be described by a <a>Thing Description</a> and are a kind of <a>Thing</a>.
```
benfrancis, line 451: This is really a "Connected Device", but "Device" is a useful shorthand. I suggest adding both terms to this definition since both terms are used in the body of the text.

index.html
```diff
@@ -716,8 +748,10 @@ <h1>Terminology</h1>
       <dt>
         <dfn>Virtual Thing</dfn>
       </dt>
-      <dd>An instance of a Thing that represents a Thing that is located
-        on another system component.</dd>
+      <dd>A <a>Service</a> that represents another <a>Thing</a>.
+       A Virtual Thing will often as as an <a>Intermediary</a>.
+       Examples include <a>Shadows</a> and <a>Digital Twins</a>.
```
benfrancis, line 753: I too have previously used the term "Virtual Thing" to mean a Web Thing which has no physical device associated with it. E.g. WebThings Gateway has a "Virtual Things add-on" which lets you add all kinds of virtual devices to your gateway which don't exist in the real world. This is useful for testing, but also for use cases like creating a virtual on/off switch which controls multiple physical devices for example.

What would be a good term for describing a Web Thing which has no physical counterpart? Could this be another kind of Virtual Thing?

benfrancis (approved):
A couple of small nits, but this is really helpful. I found the definitions of Shadow and Digital Twin to be particularly helpful because I've always struggled with the difference between a Web Thing and a Digital Twin.

I've been reluctant so far to accept the idea that any "service" can be a "thing" because it may create the perception that the Thing Description can and should be used to describe any kind of web service (like OpenAPI), and that could lead to significant scope creep. However, I think the `DirectoryThing` in WoT Discovery makes it hard to continue to argue that point since that is clearly a web service, which isn't a physical device, being described by a Thing Description.

Thumbs up from me 👍

index.html
mlagally: Arch call on 17.2.: Web Device?
Review usages of connected device?

index.html
```diff
@@ -631,13 +644,32 @@ <h1>Terminology</h1>
...
       <dd>A software stack that implements the WoT building
         blocks. A Servient can host and expose Things and/or host Consumers that consume Things.
         Servients can support multiple Protocol Bindings to enable
         interaction with different IoT platforms.</dd>
+      <dt>
+        <dfn>Shadow</dfn>
+      </dt>
+      <dd>A Shadow is a <a>Virtual Thing</a> that
+      maintains a copy of the state and mediates interactions with another
+      <a>Thing</a> which may not be immediately available due to network
```
mlagally, line 666: I suggest to define a shadow without statements about whether the thing is available or not.
IMHO a shadow reflects the state of another thing.

index.html
```diff
@@ -716,8 +748,10 @@ <h1>Terminology</h1>
       <dt>
         <dfn>Virtual Thing</dfn>
       </dt>
-      <dd>An instance of a Thing that represents a Thing that is located
-        on another system component.</dd>
+      <dd>A <a>Service</a> that represents another <a>Thing</a>.
```
mlagally, line 751: or thing*s*?

index.html
```diff
@@ -631,13 +644,32 @@ <h1>Terminology</h1>
...
+      <dt>
+        <dfn>Shadow</dfn>
+      </dt>
+      <dd>A Shadow is a <a>Virtual Thing</a> that
+      maintains a copy of the state and mediates interactions with another
+      <a>Thing</a> which may not be immediately available due to network
+      connectivity, power savings modes, or other factors.
+      A Shadow aims to achive eventual consistency with the state
+      of the Thing it represents.
+      If a Shadow has more advanced functionality than simply caching state
```
mlagally, line 670: What do you mean by "advanced functionality"? Could this be interactions via other (= non-property) affordances? 
Is a event consumer still a shadow or is it already a twin?

I believe there's no clear border between shadows and twins, suggest to use these terms as synonyms.

index.html
```diff
@@ -631,13 +644,32 @@ <h1>Terminology</h1>
       </dt>
       <dd>The combination of Public Security Metadata, Private Security Data, and any other configuration
         information (such as public keys) necessary to operationally configure the security mechanisms of a Thing.</dd>
+      <dt>
+        <dfn>Service</dfn>
```
mlagally, line 648: Should we be using the term WebService?

index.html
mlagally: I agree with @benfrancis , let's add Connected Device. We could also include Network Device as a synonym.

index.html
mmccool: hmm... you also use "virtual thing" for a "testing service" that "stands in place of" a physical thing.  I guess I could add that to the definition.  In this case, there is still "another thing" that the virtual thing is "standing in for".

index.html
mmccool: See above comments to Ben.    I plan to broaden the definition a bit/a lot and include this.  But this does not mean it's only a collection: a Virtual Thing is still a Service, so it's still a software entity with a network interface.  But yeah, there are use cases where we want to use a Virtual Thing to interface to multiple other Things.

My main concern right now is that my proposed extended definition of Virtual Thing is sounding more and more like "Intermediary" and I want to avoid having two terms for the same idea.  I will leave it alone for now, but options are: (1) restricting Intermediaries to one Exposer (2) making Virtual Things a kind of Intermediary, but also allowing Devices (e.g. physical Things) to be Intermediaries.  I'm leaning towards the second option.  I think it would be odd to call a physical device a "virtual" Thing.

index.html
benfrancis: I also wasn't sure about the distinction between a Shadow and a Digital Twin either, but the distinction being one-way (physical device to shadow) vs. two-way (between physical device and twin) makes a lot of sense. I haven't done a comprehensive literature review but this does seem to match the definition used by organisations like [Fraunhofer IESE](https://www.iese.fraunhofer.de/en/services/digital-twin.html), [MDPI](https://www.mdpi.com/2075-5309/11/4/151/pdf) and [PSA](https://www.psa.inc/company/news/digital-model-vs-digital-shadow-vs-digital-twin-and-their-place-in-eiot/).

mmccool: I would prefer to have a Shadow JUST mirror state, and be specific about that.  It's a lot easier to reason about Shadows as a part of a system if that's all it does, e.g. for the case of a hub having to deal with client devices that need to sleep and just call in occasionally on their own schedule to sync up.  If it's a Shadow we can reason about it like a slow version of the actual device.  If it's a Digital Twin it could be doing anything.
 
We could make Shadow a special case of a Digital Twin but then it's unclear whether we JUST mean mirroring when we say Digital Twin.  I would prefer to reserve Digital Twin for cases where it does more than just mirroring.

index.html
benfrancis: Just to note that since the terms "Thing" and "Web Thing" have been used as synonyms up until this point [1], giving the two terms separate definitions may create confusion and would require a full review of all specifications to make sure the terms are being used correctly.

Does it really make sense to say that a Thing that isn't a Web Thing can be described by a Web of Things Thing Description?

I won't rehash my arguments for why "MQTT Web Thing" makes no sense ;)

> I suppose you can say "HTTP Web Thing" if necessary (and maybe this is a good name for the Profile we are working on...).

I wouldn't object to that :)

--- 

I have personally often used "Thing" to describe the physical device in the real world represented on the web by a "Web Thing", but that is probably not the common understanding. Those concepts would probably be described as "Device" and "Web Thing" respectively in this latest terminology.

index.html
mmccool: One more minor technical point: hubs may have Things that "represent" other devices that may not actually have Thing Descriptions, for example a Service in a hub may present a network interface and shadow some other client device without a Thing Description, like a Z Wave or Zigbee device.  Not sure how to deal with that right now, we may have to loosen the definitions slightly to allow for shadowing of entities that don't have actual Thing Descriptions and/or that Thing Descriptions can't currently handle.

index.html
mmccool: Anyways, as I said elsewhere, that is also why I hesitate to put "Web" in front of "Service", since it may imply use of HTTP for some people.  So for now I will also leave that alone.

index.html
benfrancis: > In this case, there is still "another thing" that the virtual thing is "standing in for".

Not necessarily, it could be a virtual instance of a device which doesn't have a physical counterpart. As I said...

>> but also for use cases like creating a virtual on/off switch which controls multiple physical devices for example.

It still has all the properties of a Web Thing (a Thing Description, interaction affordances with network interfaces), it just doesn't represent an actual physical object in the real world. As far as a Consumer of a Web Thing is concerned the distinction is irrelevant, it doesn't know or care whether changes to the Web Thing manifest in the physical world.

index.html
mmccool: Re one-way vs two-way: I will have to look at the definitions you cite e.g. by Fraunhofer.  But to me the sync would go both ways.  Suppose I have a switch that for some weird reason is only a client and only syncs periodically.  The I want to "read" a binary property on the Shadow to know what state the switch is in, but I might also want to "write" the property to flip the state.  In this case the real switch (when it calls in) should check the state on the switch and sync with it.   So then the Shadow acts like a "slow" version of the real device, with a latency (at most) equal the interval at which the client device wakes up and checks in.

Anyway, I think they are different, if not for precisely this reason.

index.html
```diff
@@ -4270,10 +4304,13 @@ <h1>Recent Specification Changes</h1>
...
     <h2 id="changes-from-recommendation-1.0">Changes from the 1.0 version of [[wot-architecture]]</h2>
     <ul>
+      <li>Add definitions for Device, Service, and Shadow.  Update definitions for
+	  Virtual Thing and Digital Twin.  Narrow definition of Virtual Thing to
+	  a Service that represents another Thing.</li>
       <li>Split Security and Privacy Considerations into separate chapters.</li>
       <li>New chapter: Lifecycle.</li>
       <li>New terminology: Thing Model.</li>
-      <li>Chapter restructuring and renaming:
+      <li>Chapter restructuring and renaming
```
mmccool, line 4313: huh, not sure what happened to the colon... will fix.

index.html
mmccool: We should review how Web Thing is used.  It may be some cleanup is needed to map (Web Thing, Thing) to (Thing, Device).

mmccool:
OK, some updates on points I think we have consensus on:
- Extended definition of Virtual Things to include interfacing to one or more other Things, etc.  See new defn for detail, but I'm trying to cover the use cases mentioned without making it TOO broad.  
- Added Connected Device, but this is a pure synonym for Device.   Bidirectionally cross-referenced.
- Updated Shadow to simplify the definition (although since I don't 100% agree we should remove the detail about that use case, I left it as a comment in the source, just in case we want to revert it or add it as a "for example").  Took out "advanced" when talking about Digital Twins.

mmccool:
So... in order to answer some questions in my head about Web Thing vs. Thing, I looked at the definition for "Thing or Web Thing" and have some concerns:
- It defines "virtual entity" as a "composition"
- It says "metadata and interfaces", that maybe should be "or" (at least)

I would prefer the very simple definition as follows:
- Thing: An entity described by a Thing Description.

I think the current definition of Thing Description is ok.  But if we just say "described by a Thing Description" then we don't have to repeat ourselves with "metadata and/or interfaces and/or links..." in the definition of Thing.  And since we say Virtual Things, Devices, and Services etc. are all a kind of Thing, the "virtual entity" and "physical entity" parts can be dropped.  Finally, "abstraction" is just confusing and I feel, unnecessary.

index.html
mmccool: In particular, if it is a software component without a network interface, it is usually called a "module" or a "library".  When I hear the word "service" I immediately assume it has a network interface.

index.html
mmccool: I added "Connected Device" but did not change the definition of Web Thing.  See discussion below.

index.html
mmccool: fixed

index.html
mmccool: I meant by "standing in for" that the "virtual instance" was standing in for a "real instance".  Anyhow, please take a look at the updated definition.

mmccool:
To clarify I'm not touching Web Thing.  We discussed this at length already and while I'm not convinced we need it, I'm ok with leaving it alone and that's the easiest option.

benfrancis:
Looks good to me, thanks.

Issue #661, opened 2021-12-23
Need to have a description of a telemetry system topology in chapter 6


mlagally:
This is a very fundamental deployment scenario and should be described

mlagally:
A placeholder editors note was added in #660

mlagally:
This also requires a terminology clarification what is meant by telemetry

PR #728, opened 2022-04-01
Draft telemetry section

mlagally:
Closes [661](https://github.com/w3c/wot-architecture/issues/661)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/pull/728.html" title="Last updated on Apr 7, 2022, 6:22 AM UTC (b0260ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/728/e30d299...b0260ac.html" title="Last updated on Apr 7, 2022, 6:22 AM UTC (b0260ac)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1222,19 +1222,36 @@
     </p>
 
     <section id="telemetry">
-      <h3>Telemetry</h3>    
-      <section class="ednote" title="TODO">
-      <h3>TODO</h3>
-      Create a description and a diagram for a telemetry scenario.
-      </section>
+      <h3>Telemetry</h3>  
+      <p>
+        Telemetry is the monitoring of remove devices, which implies automatic transmission to, and processing
+        of measurements and other data on a <a>Consumer</a>. 
+        Data may be transported over various bearers, both wireless and wired.
+        Examples include GSM networks, Bluetooth, WiFi, Ethernet, and other wired standards.
+      </p>
+      <p>
+        Remote metering devices include one or multiple sensors, in some cases they also contain an actuator.
+        In many cases the sensor data is transmitted at regular intervals, on a state change, 
+        or as a response to a request by the <a>Consumer</a>.
+        Remote metering devices are frequently small embedded systems with very limited resources. 
+        These may be battery powered and have to minimize power consumption through various measures,
+        e.g. sleep modes.
+      </p>
+      <p>
+        Typical user scenarios are monitoring of a high number of geographically distributed assets, 
+        examples include smart cities, factories, fleets, environmental monitoring and health monitoring. 
+        A common pattern is the one-to-many relationship between a single <a>Consumer</a> and the devices.
+      </p>
     </section>
 
     <section id="device-controllers">
       <h3>Device Controllers</h3>
       <p>
-        The first use case is a local device controlled by a
-        user-operated remote controller as depicted in <a href="#smart-home-device"></a>. A remote
-        controller can access an electronic appliance
+        A common deployment pattern is a local device controlled by a
+        user-operated remote controller as depicted in <a href="#smart-home-device"></a>. 
+      </p>
+      <p>
+        A remote controller can access an electronic appliance
         through the local home network directly. In this
         case, the remote controller can be implemented by a
         browser or native application.
@@ -3955,7 +3972,7 @@
         <h5>Thing Description Private Security Data Risk</h5>
         <p>
           TDs are designed to carry only <a>Public Security Metadata</a>.
-          Producers of TDs must ensure that no <a>Private
+          <a>Producers</a> of TDs must ensure that no <a>Private
             Security Data</a> is included in <a>TDs</a>.
           There should be a strict separation of
           <a>Public Security Metadata</a> and <a>Private Security Data</a>.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1222,19 +1222,36 @@ <h1>Common Deployment Patterns</h1>
...
-      <h3>Telemetry</h3>    
-      <section class="ednote" title="TODO">
-      <h3>TODO</h3>
-      Create a description and a diagram for a telemetry scenario.
-      </section>
+      <h3>Telemetry</h3>  
+      <p>
+        Telemetry is the monitoring of remove devices, which implies automatic transmission to, and processing
+        of measurements and other data on a <a>Consumer</a>. 
+        Data may be transported over various bearers, both wireless and wired.
```
egekorkan, line 1229: bearer is not defined in the terminology

index.html
```diff
@@ -1222,19 +1222,36 @@ <h1>Common Deployment Patterns</h1>
...
+        In many cases the sensor data is transmitted at regular intervals, on a state change, 
+        or as a response to a request by the <a>Consumer</a>.
+        Remote metering devices are frequently small embedded systems with very limited resources. 
+        These may be battery powered and have to minimize power consumption through various measures,
+        e.g. sleep modes.
+      </p>
+      <p>
+        Typical user scenarios are monitoring of a high number of geographically distributed assets, 
+        examples include smart cities, factories, fleets, environmental monitoring and health monitoring. 
+        A common pattern is the one-to-many relationship between a single <a>Consumer</a> and a large number of devices.
```
egekorkan, line 1243: Even though I agree that this is one typical scenario, telemetry is not only this right? Usual people monitoring the energy usage of their home is also a typical scenario and there is no high number of geographically distributed assets.

egekorkan (changes requested):
request change is mostly due to my first comment about bearer

relu91 (approved):

index.html
```diff
@@ -1222,19 +1222,36 @@ <h1>Common Deployment Patterns</h1>
...
 
     <section id="telemetry">
-      <h3>Telemetry</h3>    
-      <section class="ednote" title="TODO">
-      <h3>TODO</h3>
-      Create a description and a diagram for a telemetry scenario.
-      </section>
+      <h3>Telemetry</h3>  
+      <p>
+        Telemetry is the monitoring of remove devices, which implies automatic transmission to, and processing
```
k-toumura, line 1227: remove -> remote ?

index.html
sebastiankb: I think it is also important to point out that there is a relationship to the definition of _WoT events_ that can be associated with telemetry based scenarios.

index.html
```diff
@@ -1222,19 +1222,36 @@ <h1>Common Deployment Patterns</h1>
...
     </section>
 
     <section id="device-controllers">
       <h3>Device Controllers</h3>
       <p>
-        The first use case is a local device controlled by a
-        user-operated remote controller as depicted in <a href="#smart-home-device"></a>. A remote
-        controller can access an electronic appliance
+        A common deployment pattern is a local device controlled by a
+        user-operated remote controller as depicted in <a href="#smart-home-device"></a>.
```
sebastiankb, line 1251: how about industry-based controllers?

index.html
```diff
@@ -1222,19 +1222,36 @@ <h1>Common Deployment Patterns</h1>
...
     <section id="telemetry">
-      <h3>Telemetry</h3>    
-      <section class="ednote" title="TODO">
-      <h3>TODO</h3>
-      Create a description and a diagram for a telemetry scenario.
-      </section>
+      <h3>Telemetry</h3>  
+      <p>
+        Telemetry is the monitoring of remove devices, which implies automatic transmission to, and processing
+        of measurements and other data on a <a>Consumer</a>.
```
k-toumura, line 1228: Consumer: Since the definition of WoT entities (Thing/Intermediary/Consumer) is given later in Chapter 6, so what about to use a generic term (e.g. "data collector") rather than a WoT-specific term here?

chachamimm (commented):
If comments of other members is supported, I will approve.

index.html
mlagally: This would be another example of Device Controllers.
Note that this MR focuses on the Telemetry description and just changes the numbering of the first sentence of the "Device Controllers" section.

Please create a MR to add "industrial controllers" to that section.

index.html
mlagally: @egekorkan The home energy monitoring I would have seen in the "Device Controllers" section.

index.html
mlagally: @ege We could use "communication channels" as an alternative wording.

index.html
mlagally: 👍

index.html
mlagally: @k-toumura If we would use "data collector" we would have to define it in the terminology.
A "data collector" sounds like it is only storing the data. Typically the consumer does processing, applies rules, ...

index.html
egekorkan: I think that would be better :)

index.html
mlagally: @egekorkan, @sebastiankb Thanks for your feedback.
I have updated the section to address your comments.

index.html
mlagally: We could use "data consumer" if it makes the text clearer, what do you think?

PR #732, opened 2022-04-07
Implementation Report Updates

mmccool:
- Update README
- Copy in templates for testimonials and implementation descriptions
- Get rid of some leftover TD content in categories.csv and testspec.html
- Rebase to catch up with latest revision of spec
- Regenerate report (also updates template.csv)
- Rerun extractFile.js and update assertions.csv


mmccool:
Would like to merge this shortly, @mlagally please review.

PR #733, opened 2022-04-07
Fix Producer ref and respec error

mmccool:
- Producer was, in fact, already linked from Security and Privacy, so removed ignore-lint from its dfn
- There was another ReSpec error caused by the latest update to the use case description, a simple missing period, so I fixed that while I was at it

This PR has 0 ReSpec errors.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/733.html" title="Last updated on Apr 7, 2022, 12:09 PM UTC (e5761bd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/733/7da670c...mmccool:e5761bd.html" title="Last updated on Apr 7, 2022, 12:09 PM UTC (e5761bd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -315,7 +315,7 @@
       <li>Domain specific (vertical) use cases for a single application domain</li>
     </ul>
     <p>These use cases and requirements drive the creation and further evolution
-      of the W3C WoT specification family</p>
+      of the W3C WoT specification family.</p>
 
     <p>The WoT architecture specification is focused on the scope of W3C WoT standardization,
       which can be broken down into these building blocks as well as the abstract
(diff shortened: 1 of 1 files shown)


mmccool:
This is a very very minor bug-fix so I would suggest it can be merged without waiting for the next arch call, if the reviewer agrees.

mmccool:
I note in passing though that there are in fact a lot of other terms with lint-ignore on them that we do in fact use, like "Servient".  Probably we should remove all the lint-ignore tags and try to use all the terms we define somewhere...  although some terms like "Security" are used so much that it would be a nuisance, but we could link each definition at least once.

Issue #672, opened 2022-01-10
Review Security and Privacy Considerations


mmccool:
The security and privacy considerations section needs to be reviewed and updated. 

See also similar issues for other deliverables; we need to also decide what goes where:
* https://github.com/w3c/wot-thing-description/issues/1348
* https://github.com/w3c/wot-discovery/issues/254

mmccool:
One possible security consideration to add would be to address the issue raised in https://github.com/w3c/wot-security/issues/197, that WoT tends to encourage making all Things servers but that servers are harder to secure than clients.

mmccool:
Notes:
- Overlap with other docs, e.g. TD.  These are MEANT to be summaries, but...
- No Discovery security and privacy considerations 
- No Profile security and privacy considerations
One simple fix would be to take out any considerations that relate to another spec and deal with them in that spec, and put only considerations that are not covered elsewhere in Arch.

mmccool:
The WIP PR #679 was merged and closed which also closed this issue, but it has not actually been resolved yet.

mmccool:
Section 6, "System Topologies", has numerous diagrams that use "Trusted Environments" but this is never defined.  In addition to a Security Consideration describing what this means in detail and related topics like segmented networks, zero trust, etc. I think adding a short definition to the opening paragraph and pointing at the relevant security considerations section is appropriate.

mmccool:
An issue raised in the TD call also applies here: https://github.com/w3c/wot-thing-description/issues/1363
Apparently wide review (and respec) now require separate "Security Considerations" and "Privacy Considerations".  So my PR now does this too (or has started to; it also requires cleaning up a bunch of references, how we talk about Security and Privacy being cross-cutting issues, etc. etc. and is basically a pain).

PR #734, opened 2022-04-07
Make Security and Privacy Considerations Normative

mmccool:
addresses this point in issue #672 - makes consistent with other deliverables (Discovery and TD 1.1)

- mark both Security Considerations and Privacy Considerations sections as normative
- update change log
- remove "refactoring" ed note (actually resolved a long time ago)
- add assertion spans and RFC2119 keywords to normative mitigations
- some rewording needed to make the above possible in some cases

Given the timing and the cancellation of Architecture calls over the next two weeks, we will review and revise these in the Security TF and based on PR reviews aiming for a final version by end of April so we can hit the normative feature freeze deadline of May 6.  Unfortunately some of the text does needs revision for the assertions to make sense as standalone sentences and not be redundant, and to follow the format (informative description of risk, followed by normative mitigations; in some cases, the mitigations and risks were mixed together, resulting in assertions in the "risk description" section, and that needs to be sorted out).

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/734.html" title="Last updated on Apr 28, 2022, 12:49 AM UTC (be047ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/734/7da670c...mmccool:be047ef.html" title="Last updated on Apr 28, 2022, 12:49 AM UTC (be047ef)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3917,13 +3917,8 @@
     </section>
   </section>
 
-  <section id="sec-security-considerations" class="informative">
+  <section id="sec-security-considerations">
     <h1>Security Considerations</h1>
-    <p class="ednote">
-        Need to be updated to include points noted in <a href="https://github.com/w3c/wot-architecture/issues/672">Architecture Issue 672</a>.
-	Also needs to be refactored into separate sections, which is going to require
-	rewriting the intro (see the next section).  Also the ids and references all need to be updated...
-    </p>
     <p>
       Security is a cross-cutting issue that needs to be considered
       in all <a href="#sec-building-blocks">WoT
@@ -4240,7 +4235,7 @@
       </dl>
     </section>
   </section>
-  <section id="sec-privacy-considerations" class="informative">
+  <section id="sec-privacy-considerations">
     <h1>Privacy Considerations</h1>
     <p>
       Privacy is a cross-cutting issue that needs to be considered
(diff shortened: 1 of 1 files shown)


k-toumura:
> I'd like to propose flattening the structure and instead of putting considerations under sections associated with each deliverable,
I agree.  The Architecture document is suitable for describe considerations that spans multiple deliverables, so it is would be better to flatten the structure and add Impacted Deliverable section.

mmccool:
Considered "flattening" these to avoid putting particular risk under particular deliverables but this turned out to be awkward.  I settled for (a) moving all assertions under mitigations where they belong (b) mentioning other programmable systems besides the scripting API can be subject to the risks mentioned there (c) mentioning discovery as a mechanism designed to mitigate the TD distribution risks.

I consider this ready to merge myself, but we can discuss whether another week of review would be useful (it hasn't changed much with my latest edits, and otherwise has been stable for several weeks).

Citrullin:
@mmccool "Configure any communication security metadata for an IoT Platform used with WoT following best practices for that IoT Platform." This quote is a little hard to understand.
How about: "Any communcation security metadata for an IoT platform, used with WoT, should follow the best practices of the given IoT Platform."

The rest looks fine to me.

Citrullin:
>  Even if required by the use case, to minimize tracking risks, distributed and limited-scope identifiers should be used whenever possible rather than globally unique identifiers

Maybe you can make it actually a SHOULD instead.

PR #747, opened 2022-05-09
Additional Security/Privacy Considerations around TLS, access controls for PII

mmccool:
This is an attempt to address the issue of when and where TLS and access controls are required.  Initial feedback from PING is requesting that we make access controls mandatory when PII is involved, but this is only effective if there is also secure transport, which we can't mandate in a blanket fashion due to limitations on deploying it on a LAN.

In summary, this PR adds the following rules:
- Secure transport (TLS) mandatory for public internet access.  
- Secure transport (TLS) recommended for private networks (e.g. LANs), but optional.
- Access controls mandatory for access to PII by people other than the person the PII is about.  This includes TDs describing Things related to people in directories, but also data from normal Things associated with a person.

### DIFF
--- a/index.html
+++ b/index.html
@@ -499,8 +499,12 @@
       <dt>
         <dfn data-lt="WoT Discoverer" class="lint-ignore">Discoverer</dfn>
       </dt>
-      <dd>An entity that generates and registers a TD with an exploration service
-        on behalf of a <a>Thing</a> which may not be able to do it for itself.
+      <dd>An entity which acts as a client of a WoT Discovery service to discover 
+        and fetch a <a>Thing Description</a>, 
+        e.g. by being introduced to and searching a 
+        <a>Thing Description Directory</a> exploration service or by
+        fetching a <a>Thing Description</a> directly from 
+        the well-known endpoint on a Thing.
       </dd>
       <dt>
         <dfn>Domain-specific Vocabulary</dfn>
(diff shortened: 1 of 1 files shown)


mmccool:
Argh, I accidentally put the Discoverer and Registrant defns in my main, then branched off that... pls ignore those for now, if the other PR containing them does not get merged I will remove them from this one, and otherwise will rebase.

mmccool:
The assertion "Things returning data associated with a person accessible to anyone other than that person MUST use some form of access control." should probably be changed to "Things returning data associated with a person or with an immutable identifier MUST use some form of access control."

This is because the condition "except by that person" can only be checked by having some kind of access control.  The "immutable ID" is meant to address the case where an ID cannot be changed by law, e.g. for medical device.  See the PING issue.

I also put this in Architecture even though the PING comment was on TDs since it also applies to Discovery and I think putting it in a central place it better.  I probably should cite it or something in the TD and Discovery specifications (and/or modify the appropriate assertions in those specs, e.g. assertions about use of TLS in Discovery, see Ben's review comments and PRs addressing it)

mmccool:
Right now says "MAY" use Trusted Environment on a LAN.  To be stronger, even though it is logically redundant, I probably should add a "SHOULD" assertion for private networks paralleling the MUST assertion for public networks.

mmccool:
Also, the "SHOULD" for "private networks" should have an informative statement that "The risk increases with the sensitivity of the data being protected and with the number of people given access to the private network.  Secure transport is a higher priority in higher risk situations such as a factory network or a university campus; in such cases, it is  also more practical to install pre-shared keys."

mmccool:
as discussed, fixed a few small things:
- Discovery process (align with other PR)
- Fix typo, Unfor(t)unately
- Add missing close section tag

PR #769, opened 2022-05-19
Internationalization Questionnaire Answers

mmccool:
- i18n self-checklist needed prior to wide review 
- mostly defers to other deliverables but does summarize issues
- also removes obsolete (misspelled filename) copy of accessibility questionnaire


mlagally (approved):
Thanks for the accessibility cleanup.

PR #783, opened 2022-06-23
Specify TLS and DTLS versions

### LINKED ISSUE #753: Clean up new TLS S&P Considerations
Recently merged [PR #747](https://github.com/w3c/wot-architecture/pull/747) regarding use of TLS, but...

- reference for TLS would be good
- mention version of TLS?  Of just "up-to-date version"
- mention DTLS as suitable in case of UDP-based protocols
- Two uses of "generally" in the same sentence - needs english cleanup

Please comment below if you see anything else that needs fixing.

mmccool:
Resolves #753 

- remove two generallys
- SHOULD use TLS 1.3
- MAY use TLS 1.2
- SHOULD use DTLS 1.3
- MAY use TLS 1.2
- MUST not use earlier versions of TLS or DTLS for new development (as opposed to describing brownfield devices)
- RFC references for the above

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/783.html" title="Last updated on Jun 30, 2022, 10:01 AM UTC (77a9070)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/783/af14f99...mmccool:77a9070.html" title="Last updated on Jun 30, 2022, 10:01 AM UTC (77a9070)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4330,10 +4330,10 @@
       there are situations in which secure transports, such as TLS, are not
       feasible or are difficult to set up, such as on a local LAN within a home.
       Unfortunately, generally access control mechanisms in HTTP 
-      are generally designed to be used with secure transport and can be 
+      are designed to be used with secure transport and can be 
       easily bypassed without it.  In particular, it is relatively easy to
       capture passwords and tokens from unencrypted protocol interactions
-      which can be intercepted by a third party.  In addition, man-in-the-middle
+      which can be intercepted by a third party.  Also, man-in-the-middle
       attacks can be easily implemented without TLS providing server authentication. 
       </p>
       <dl>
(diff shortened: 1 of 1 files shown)


mmccool:
Discussion in Security TF call 27 June:
- Existing application and devices might use TLS 1.1, and we can "describe" these, but they are not secure.
- Could add assertions such as "Except for the description of existing devices and applications that cannot be upgraded, the original version of DTLS and TLS 1.1 and earlier MUST NOT be used."

mmccool:
Updated, some tweaks (due to annoying problems referencing older versions of DTLS and TLS, but also SSL, etc) added the following:
```html
<span class="rfc2119-assertion"
              id="arch-security-consideration-no-earlier-tls-or-dtls">
            Versions of DLTS or TLS earlier than 1.2 MUST NOT be used for
            new development.
          </span>
          Existing Thing using earlier versions of TLS or DTLS can be 
          described by WoT metadata (e.g. Thing Descriptions) but should
          be considered insecure.
```

PR #788, opened 2022-07-03
Editorial fixes for section 4 + 5

mlagally:


index.html
```diff
@@ -1353,14 +1354,14 @@ <h3>Thing-to-Thing</h3>
         other to actuate or notify.</p>
       <figure id="smart-home-t2t">
         <img src="images/wot-use-cases/smart-home-t2t.png" srcset="images/wot-use-cases/smart-home-t2t.svg"
-          class="wot-arch-diagram" alt="smart home t2t use case" />
+          class="wot-arch-diagram" alt="smart home t2t deployment scenario" />
```
k-toumura, line 1357: Since the title of Chapter 5 is "Common Deployment Patterns", how about "deployment pattern" instead of "deployment scenario"?

(The same applies to other parts where "use case" is changed to "deployment scenario".)

mlagally: Arch call on July 14th: we created a new issue for that #803

PR #827, opened 2022-09-01
remove arch-security-consideration-public-metadata-only; redundant

### LINKED ISSUE #824: Adjust policy-like security and privacy assertions
Action for item 4 in https://github.com/w3ctag/design-reviews/issues/736#issuecomment-1187325252 for Architecture.
Overlaps with other issues (todo, add links here) that will look at entire document.  However, this issue will document issues in Security Considerations and Privacy Considerations that may need adjusting specifically, after review by the Security TF.

Note: we want to remove (or modify) assertions that specifically can only be asserted about *deployments*, not implementations.  However, there are many assertions in architecture that can be read as *requirements* for other specifications, such as TD.  To the extent that those specifications satisfy the requirement, we can consider them true.  In other words, if they are *verifiable* at the point of implementation, we can keep them.

Assertions that need adjusting:
* [arch-security-consideration-no-private-security-data](https://w3c.github.io/wot-architecture/#arch-security-consideration-no-private-security-data) - redundant with [arch-security-consideration-public-metadata-only](https://w3c.github.io/wot-architecture#arch-security-consideration-public-metadata-only), and does not mention extensions.
* [arch-security-consideration-other-programming-mechanisms](https://w3c.github.io/wot-architecture#arch-security-consideration-other-programming-mechanisms) - not inside a mitigation, too general, does not use RFC-2119 keywords: turn into informative text
* [arch-security-consideration-isolation-sensitive](https://w3c.github.io/wot-architecture#arch-security-consideration-isolation-sensitive) and [arch-security-consideration-isolation-tenants](https://w3c.github.io/wot-architecture#arch-security-consideration-isolation-tenants) both require strict isolation of WoT runtimes, but while this would be a very useful property is difficult to achieve and goes beyond the requirements for a runtime, but rather affects how a runtime is instantiated (eg. in secure containers, etc).  So it's a deployment issue, not an implementation issue, per se.  Also perfect isolation is still hard to achieve with current technology, so probably asking for too much.  Downgrade to informative statements.

10.2.2, 10.3.1: keep assertions, conceptually about an implementation, but will need manual assertions. 

10.3.2: "should guarantee isolation of" -> "SHOULD isolate"

10.4: should not be a problem to validate.  Note however that static IPs may be necessary to set up certs on LANs.  Also applies to 10.5.  Assertion for "implicit access control via access to a common network" should be changed to a MUST.

10.5: ok as is

11.1.1: may overlap some with assertions in TD spec, but for now will just look at these in-place.  "this specific issue" -> "these specific issues".  Last assertion should be changed to "As a matter of policy, ..." and made informative.

11.2: verifiable, but may overlap with some assertions in the TD spec.  Check for overlap (and conflict) with TD assertions. 

The following are requirements of specifications, and not assertions about implementations per se.  We MAY want to find another way to express these rather than using RFC-2119 Assertions.

mmccool:
resolves #824 

- remove arch-security-consideration-public-metadata-only; redundant.  Could have removed either assertion about public or private metadata, but assertion about not including private metadata is more important.  Also reordered assertions in this section.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4104,10 +4104,6 @@
 	  <span class="rfc2119-assertion" id="arch-security-consideration-separate-security-data">
           There SHOULD be a strict separation of
           <a>Public Security Metadata</a> and <a>Private Security Data</a>.</span>
-	  <span class="rfc2119-assertion" id="arch-security-consideration-public-metadata-only">
-	  <a>Producers</a> of TDs and extensions meant to be used in TDs
-          MUST ensure that only <a>Public Security Metadata</a>
-          is ever stored in TDs.</span>
 	  <span class="rfc2119-assertion" id="arch-security-consideration-auth-private-data">
 	  Authentication and authorization 
 	  SHOULD be established based on separately managed <a>Private Security Data</a>.</span>


index.html
```diff
@@ -4571,12 +4566,13 @@ <h5>Thing Description Personally Identifiable
...
+	    Note that the <a>WoT Discovery</a> mechanism is designed to address these
+	    specific issues, as long as it is used with authentication and access
 	    controls on exploration services.
-	    <span class="rfc2119-assertion" id="arch-privacy-consideration-no-unness-info-td"> 
-            Unnecessary information
-            SHOULD NOT be exposed in TDs whenever possible.</span>
+	    <!--span class="rfc2119-assertion" id="arch-privacy-consideration-no-unness-info-td" -->
+      As a general matter of policy,
+            unnecessary information
+            should not be exposed in TDs whenever possible.<!-- </span> -->
```
mlagally, line 4575: uppercase SHOULD

index.html
```diff
@@ -4605,25 +4601,29 @@ <h2>Access to Personally Identifiable Information</h2>
...
-      required to use access control if the TDs describe Things associated with
+      should use access control if the TDs describe Things associated with
       identifiable people.  In the case of services
       returning Thing Descriptions, the following also applies: 
 	      <span class="rfc2119-assertion" id="arch-privacy-consideration-id-access-control-mandatory-immutable">
-Services returning Thing Descriptions with immutable IDs MUST use some form of access control.</span>
+Services returning Thing Descriptions with immutable IDs SHOULD use some form of access control.</span>
       Specifically, in both of these situations, the <code>nosec</code> security
-      scheme described in [[WOT-THING-DESCRIPTION]] cannot be used,
+      scheme described in [[WOT-THING-DESCRIPTION]] should not be used,
```
mlagally, line 4615: assertion?

index.html
```diff
@@ -4187,27 +4182,26 @@ <h5>Cross-Script Security Risk</h5>
...
-	    Such isolation
-            can be performed within the <a>WoT Runtime</a>
-            using platform security mechanisms available on
-            the device. For more information see Sections
-            "WoT Servient Single-Tenant" and "WoT Servient
-            Multi-Tenant" of the <em>WoT Security and Privacy
-              Guidelines</em> specification [[WOT-SECURITY]].
+      In practice, isolation of scripts and runtime instances from each other
+      can be accomplished by running all instances
+      in a "sandboxed" environment that controls its access to the rest of the environment.
```
mlagally, line 4202: "sandboxed environment" would benefit from an entry in terminology

mlagally (changes requested):

Citrullin:
A lot of changes make sense to me. Especially considering current IoT environments without any security. We have to move away from it, but it makes sense to allow them to not have the environment, when having good arguments to do so.

PR #839, opened 2022-09-14
add initial manual.csv for node-wot

danielpeintner:
Note: I have doubts in many places. @relu91 @egekorkan can you check also. Thanks!

BTW, most of the "arch-security-consideration-xxx " assertions are simply left blank since this depends on how it is used (and is not enforced by node-wot). That's at least how I see it. Please comment.


danielpeintner:
replaced by https://github.com/w3c/wot-testing/pull/405

PR #848, opened 2022-09-29
Update Implementation Report Oct 2022

mmccool:
- with node-wot results


danielpeintner:
see https://github.com/w3c/wot-testing/pull/412
I used "impl" instead of "pass" 🙈

PR #855, opened 2022-10-18
Clarify interaction affordance binding mechanisms  

### LINKED ISSUE #850: Contradiction with the WoT Profile spec on protocol bindings
> [Protocol Bindings](https://w3c.github.io/wot-architecture/#dfn-wot-protocol-binding) MUST be serialized as [hypermedia controls](https://w3c.github.io/wot-architecture/#sec-hypermedia-controls) to be self-descriptive on how to activate the [Interaction Affordance](https://w3c.github.io/wot-architecture/#dfn-interaction-affordance).

I suggest changing this assertion to reflect the fact that protocol bindings can also be defined in profiles.

See also:
- https://github.com/w3c/wot-profile/issues/299 (WoT Profile)
- https://github.com/w3c/wot-thing-description/issues/1674 (WoT Thing Description)

mlagally:
resolves #850

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/pull/855.html" title="Last updated on Oct 18, 2022, 3:14 PM UTC (9a0169e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/855/efb4cf6...9a0169e.html" title="Last updated on Oct 18, 2022, 3:14 PM UTC (9a0169e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2554,8 +2554,9 @@
           <span class="rfc2119-assertion" id="arch-hypermedia">
             Interaction Affordances MUST include one or more Protocol Bindings.</span>
           <span class="rfc2119-assertion" id="arch-hypermedia-protocol-binding">
-            <a>Protocol Bindings</a> MUST be serialized as <a href="#sec-hypermedia-controls">hypermedia controls</a> to be self-descriptive on how to
-            activate the <a>Interaction Affordance</a>.</span>
+            <a>Protocol Bindings</a> MUST be serialized as <a href="#sec-hypermedia-controls">hypermedia controls</a> 
+		 to be self-descriptive on how to
+            activate the <a>Interaction Affordance</a> or defined in a <a>Profile</a>.</span>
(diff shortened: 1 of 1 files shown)


egekorkan:
A couple of comments here:
- It is still not clear what a profile can contain. Thus, it is unsafe to make this assertion.
- Profile will not be a REC when this document is published so there is no way to explain what this can mean to readers
- As I have mentioned before, I do not think that implementors read the arch spec while doing their implementation so this point is not relevant but isn't this a change that is "a bit" backwards incompatible? This is the reason why node-wot has architectural problems with profiles.
- Allowing this goes against the hypermedia controls design of the WoT

chachamimm (approved):

mlagally:
Arch call on Nov 2nd.: Agree to close without merging

PR #856, opened 2022-10-18
Additional terminology entry for Web Object

mlagally:
Additional terminology entry for Web Object, which is an entity with a network interface.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/pull/856.html" title="Last updated on Oct 19, 2022, 10:27 AM UTC (5eee24a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/856/efb4cf6...5eee24a.html" title="Last updated on Oct 19, 2022, 10:27 AM UTC (5eee24a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -892,15 +892,15 @@
         form and not as full IRIs.
       </dd>
       <dt>
+        <dfn class="lint-ignore">Web Object</dfn>
+      </dt>
+      <dd>An entity that has a network interface. 
+	Examples include devices, directories, consumers and things.</dd>
+      <dt>
         <dfn>WoT Interface</dfn>
       </dt>
       <dd>The network-facing interface of a Thing
         that is described by a WoT Thing Description.</dd>
-      <dt>
-        <dfn>Web Object</dfn>
-      </dt>
-      <dd>An entity that has a network interface. 
-	Examples include devices, directories, consumers and things.</dd>
       <dt>
         <dfn>WoT Runtime</dfn>
       </dt>


mmccool:
Is this to capture existing usage in the document?   Also, how does it relate to other terms like Service and Virtual Thing?  I assume the intention is to include services (since Directories are mentioned) but it would be good to be explicit.  I'm not quite sure how this fits into the ontology or how it is different from (or a synonym for) other definitions we have.  Using actual links to other defined terms (and using caps for them) would also be good.

Also, it should be noted that "network interface" is a little vague, since an entity could have a network interface (e.g. an ethernet jack and hardware) but not a network API (e.g. application-level protocol with defined endpoints) and could also be using a protocol like MQTT or OPC UA (or MODBUS-TCP if you wanted an outlier).  I am personally fine with this term including non-HTTP protocols but I wanted to confirm that is the intended meaning.  To tighten things a little bit we may at least want to say "network API" here instead.

Generally I am a little reluctant to add new content to the document at this point unless there is a VERY good reason...

egekorkan:
I second @mmccool 's comment. My reasons:
- I do not understand why this is needed in the first place. More specifically, what issue is addressed?
- Where did we agree that this is the term to describe this definition? Like @mmccool above, why is this different than a Servient?
- It is not a good idea to have two terms like Web Thing and Web Object , that can even mean the same thing in some languages or contexts. This would confuse readers
- Why is lint ignore added? Better to use a definition if it is defined. Why add a definition in the first place if it is not used at all?
- The PR https://github.com/w3c/wot-architecture/pull/857 does this as well and that one is just merged without any discussion, review or gives any time to review? It is created 5h ago and merged 5h ago, giving 59 minutes to review in the best case.

k-toumura:
I've previously drawn the following diagram to help organize terms such as Thing and Consumer.   I think this might be useful to clarify the meaning of "Web Object".
diagram
(https://github.com/w3c/wot-architecture/blob/main/proposals/wot-arch-terms.drawio.svg; you can edit it by https://app.diagrams.net/ )

Personally, I think it is better to postpone major architectural update to Arch2.0 as we are already in the process of preparing for CR....

mlagally:
Motivation for inclusion of Web Object:

We have several components in the Web which are not things, i.e. cannot be described with a TD. We discussed for example that we cannot describe some consumers with a TD, so this is something like an abstract base class.

See also the diagram provided by @k-toumura: The discoverer and the consumer don't have TDs. 

This is a only an additional entry in the infomative terminology section without any spec impact. It will be useful for the consumer description in the Profile.

egekorkan:
> We have several components in the Web which are not things, i.e. cannot be described with a TD. We discussed for example that we cannot describe some consumers with a TD, so this is something like an abstract base class.

How is this different than a servient in this case?

mlagally:
@k-toumura 

> I've previously drawn the following diagram to help organize terms such as Thing and Consumer. I think this might be useful to clarify the meaning of "Web Object".
> diagram
> (https://github.com/w3c/wot-architecture/blob/main/proposals/wot-arch-terms.drawio.svg; you can edit it by https://app.diagrams.net/ )

This diagram is very useful and we should have included it into the spec in the terminology section. Perhaps we still should do that as an editorial update? The Terminology section is informative.

mlagally:
Arch call on Oct 27.
No consensus, agree to keep open and revisit. This is not blocking CR
If additional terminology is required in Profile, Profile should define it there.

index.html
```diff
@@ -891,6 +891,11 @@ <h1>Terminology</h1>
         <a>Vocabulary Terms</a> present in this document are always written in a compact
         form and not as full IRIs.
       </dd>
+      <dt>
+        <dfn class="lint-ignore">Web Object</dfn>
+      </dt>
+      <dd>An entity that has a network interface. 
+	Examples include devices, directories, consumers and things.</dd>
```
mlagally, line 898: Since Web is all about HTTP, this could be refined.

PR #860, opened 2022-10-20
Clarify interaction affordance binding mechanism-2

### LINKED ISSUE #850: Contradiction with the WoT Profile spec on protocol bindings

mlagally:
This is an alternative proposal for the description of the protocol binding mechanisms. It takes into account the discussion in https://github.com/w3c/wot-architecture/issues/850#issuecomment-1282644919 and implements the approach that was proposed by @mmccool 

It resolves #850

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/pull/860.html" title="Last updated on Oct 20, 2022, 6:41 AM UTC (c37b19b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/860/efb4cf6...c37b19b.html" title="Last updated on Oct 20, 2022, 6:41 AM UTC (c37b19b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2553,9 +2553,17 @@
         <p>
           <span class="rfc2119-assertion" id="arch-hypermedia">
             Interaction Affordances MUST include one or more Protocol Bindings.</span>
-          <span class="rfc2119-assertion" id="arch-hypermedia-protocol-binding">
-            <a>Protocol Bindings</a> MUST be serialized as <a href="#sec-hypermedia-controls">hypermedia controls</a> to be self-descriptive on how to
-            activate the <a>Interaction Affordance</a>.</span>
+
+          <span class="rfc2119-assertion" id="arch-hypermedia-protocol-binding-1">
+	    <a>Protocol Bindings</a> MUST be serialized in such a way that they are self-descriptive or otherwise 
+            clearly associated with definitions that indicate how to activate the <a>Interaction Affordance</a>.
+	  </span>
+
+          <span class="rfc2119-assertion" id="arch-hypermedia-protocol-binding-2">
+	    This MAY be achieved by <a href="#sec-hypermedia-controls">hypermedia controls</a>, 
+		by using <a>Profiles</a> or by using protocol specific metadata. 
+	  </span>
+		  
           The authority of the <a>hypermedia controls</a> can be the <a>Thing</a> itself, producing the <a>TD</a> document
           at runtime (based on its current state and including network parameters such as its IP address)
           or serving it from memory with only the current network parameters inserted.


index.html
```diff
@@ -2553,9 +2553,17 @@ <h3>Hypermedia-driven</h3>
         <p>
           <span class="rfc2119-assertion" id="arch-hypermedia">
             Interaction Affordances MUST include one or more Protocol Bindings.</span>
-          <span class="rfc2119-assertion" id="arch-hypermedia-protocol-binding">
-            <a>Protocol Bindings</a> MUST be serialized as <a href="#sec-hypermedia-controls">hypermedia controls</a> to be self-descriptive on how to
-            activate the <a>Interaction Affordance</a>.</span>
+
+          <span class="rfc2119-assertion" id="arch-hypermedia-protocol-binding-1">
+	    <a>Protocol Bindings</a> MUST be serialized in such a way that they are self-descriptive or otherwise 
+            clearly associated with definitions that indicate how to activate the <a>Interaction Affordance</a>.
```
sebastiankb, line 2559: instead "clearly" I would be more stronger here by "unambiguously"

k-toumura (approved):

index.html
```diff
@@ -2553,9 +2553,17 @@ <h3>Hypermedia-driven</h3>
...
+	  </span>
+
+          <span class="rfc2119-assertion" id="arch-hypermedia-protocol-binding-2">
+	    This MAY be achieved by <a href="#sec-hypermedia-controls">hypermedia controls</a>, 
+		by using <a>Profiles</a> or by using protocol specific metadata.
```
egekorkan, line 2564: When the profiles is not published, we cannot really refer to it right? This is not a normative reference (not using [[my ref]] so it might be ok but it is still NOT clear what a profile can do. If we look at https://w3c.github.io/wot-architecture/#profile-description-methodology , it only says that it can constrain the protocol verbs etc. Thus, a profile cannot define a protocol binding by this definition.

egekorkan (changes requested):

index.html
sebastiankb: I think it's fine as long as there is a Profile definition. Using <a>Profiles</a> would expect this otherwise it will emphasize with a curled red line.

benfrancis: Since WoT Profile 1.0 will now be published after WoT Architecture 1.1 and WoT Thing Description 1.1, I have proposed moving the definition of "profile" from the Profile specification to the TD or Architecture specification in order to resolve the dangling reference - see https://github.com/w3c/wot-thing-description/issues/1719.

Also see https://github.com/w3c/wot-thing-description/issues/1674 in which I proposed an updated description of protocol bindings in the TD specification which takes profiles into account. The wording in that issue describes the two different types of protocol bindings and how they are used.

It would be simpler if there were no normative definitions in the WoT Architecture specification at all, but as long as they exist they need to be kept in sync with the other specifications, since they all cross reference each other.

chachamimm (commented):
I have two questions.

1. In this PR, a new assertion was cleated. But to add new features in wot architecture 1.1 was frozen. If new features should be added, it is important to think the reason.
2. Each Feature( or assertion)  must be implemented.  Have these assertions already been implemented or will they be implemented soon?  The end of the charter period is near. So we should pay attention to the implementation status.

mmccool:
Note that adding new assertions causes a lot of problems at this point: we have to update the IR, update the at-risk list, do testing, etc.  Ideally we would just modify an existing assertion if a normative change is needed.  Not that it's impossible, just keep in mind we have to do it (at least adding it to the list of at-risk features in the sotd section).

mlagally:
Arch call on Nov 2nd.: Agree to close without merging

PR #865, opened 2022-10-25
Adding terminology "WoT Profile"

mlagally:
Implement the proposal from TD issue #1719

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/pull/865.html" title="Last updated on Oct 25, 2022, 9:40 AM UTC (7c519c0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/865/efb4cf6...7c519c0.html" title="Last updated on Oct 25, 2022, 9:40 AM UTC (7c519c0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -700,6 +700,13 @@
       <dd>An entity that can create WoT Thing Descriptions
         for a specific Thing.</dd>
       <dt>
+        <dfn>WoT Profile</dfn>
+      </dt>
+      <dd>A technical specification which provides a set of assertions such that any <a>Consumer</a> 
+	which conforms with the those assertions is out-of-the-box interoperable with any <a>Thing</a> 
+	which also conforms with those assertions.
+      </dd>
+      <dt>
         <dfn>Property</dfn>
       </dt>
       <dd>An Interaction Affordance that exposes state of the Thing.
@@ -2925,7 +2932,7 @@
     <section id="wot-profiles">
       <h3>Profiles</h3>
       <p>
-      The WoT Profile Specification [[!WOT-PROFILE]] defines Profiles, 
+      The <a>WoT Profile</a> Specification [[!WOT-PROFILE]] defines Profiles, 
       that enable <em>out of the box interoperability</em> among things 
       and devices. Out of the box interoperability implies, 
       that devices can be integrated into various application 


index.html
```diff
@@ -699,6 +699,13 @@ <h1>Terminology</h1>
       </dt>
       <dd>An entity that can create WoT Thing Descriptions
         for a specific Thing.</dd>
+      <dt>
+        <dfn>Profile</dfn>
+      </dt>
+      <dd>A technical specification which provides a set of assertions such that any <a>Consumer</a>
```
sebastiankb, line 705: Two comments here:
- do we need another definition what "out-of-the-box" mean?
- I think, the definition can be more clear that the Profile is a subset of the TD specification by providing additional assertions. 
- Proposal: A technical specification that is a subset of the Thing Description specification which provides additional set of assertions such that any <a>Consumer</a> which conforms with the those assertions is out-of-the-box interoperable with any <a>Thing</a> which also conforms with those assertions.

chachamimm (approved):

mmccool:
I agree with Sebastian that we should keep the definition simple and not add extra embedded definitions.  In fact I would not "guarantee" oobi because it's difficult to prove.  The INTENT to improve it is enough.  May I suggest:

WoT Profile: A named subset of Thing Descriptions defined by a set of assertions.  The intent of a profile is to improve interoperability between Consumers and Things whose Thing Descriptions conform with those assertions.

index.html
benfrancis: @sebastiankb What does it mean to be a subset of a specification?

A profile is a collection of assertions which are neither a subset nor a superset of the assertions in the Thing Description specification, they are a separate collection of assertions with may define additional constraints and assumptions for conformant Consumers and Things.

I suspect what you're trying to say is that a Thing Description conforming to a profile can only use a subset of the features defined in the Thing Description specification, but that isn't really an accurate description either. A profile may define a subset of TD features (such as security schemes, protocol bindings, and semantic contexts) which a conformant Consumer must support, but it may also define additional assumptions (such as protocol binding defaults), some of which can't even be described declaratively in a Thing Description. That could be described as a superset of the features described in the TD specification.

This current definition of a profile has already been reviewed and landed in the WoT Profile specification, so whatever lands here will need to replace or be consistent with that definition.

benfrancis:
@mmccool Please see my comment at https://github.com/w3c/wot-architecture/pull/865#discussion_r1006701860

index.html
benfrancis: Note that I've started a discussion about a ["Profile Mechanism 2.0"](https://github.com/w3c/wot-profile/issues/285) which could be much more rigidly defined, but that is a different approach to profiles than is taken in the current specification.

mlagally:
Arch call on Oct 27th:
Unanimous Consensus

benfrancis:
I object to this definition, it's technically incorrect.

PR #871, opened 2022-11-02
Option 2: Define whole Chapter 7 non-normative

sebastiankb:
this is an alternative PR to #862

Since there are no normative assertions in this chapter, there is no reason to define the whole section as normative.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/pull/871.html" title="Last updated on Nov 2, 2022, 10:54 AM UTC (41f89d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/871/8cdbb9e...41f89d7.html" title="Last updated on Nov 2, 2022, 10:54 AM UTC (41f89d7)">Diff</a>

PR #872, opened 2022-11-02
Update CR and Norm/Inf for section 7

mmccool:
Unfortunately, [PR 871](https://github.com/w3c/wot-architecture/pull/871) merged previously does not lead to a valid publication since, despite what we said earlier, sections 7.4 and 7.5 DO have assertions and must be normative.  Therefore, we cannot mark all of Section 7 as being informative.

This PR does the following:
- Marks section 7 as being normative again
- Marks any subsection of section 7 that does NOT have assertions as being informative.  So only sections 7.4 and 7.5 are normative
- Updates the CR candidate document in publication/ver11/3-cr to be consistent, and generates a static version, etc.

I believe this follows the intent of the original proposal as much as possible.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/872.html" title="Last updated on Nov 2, 2022, 2:32 PM UTC (415982c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/872/1c7ba10...mmccool:415982c.html" title="Last updated on Nov 2, 2022, 2:32 PM UTC (415982c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2802,7 +2802,7 @@
     </section>
   </section>
 
-  <section id="sec-building-blocks" class="informative">
+  <section id="sec-building-blocks">
     <h1>WoT Building Blocks</h1>
 
     <p> The Web of Things (WoT) building blocks allow the
@@ -2849,7 +2849,7 @@
       be considered an additional building block.
     </p>
 
-    <section id="sec-thing-description">
+    <section id="sec-thing-description" class="informative">
       <h2>WoT Thing Description</h2>
       <p> The <a>WoT Thing Description</a> (TD) specification
         [[?WOT-THING-DESCRIPTION]] defines an <em>information model</em>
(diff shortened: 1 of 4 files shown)


mlagally:
In case we want to revisit the main call resolution, I think we should simply revert the merge of PR#871 and add the Profile spec to the „at risk“ items. 

The at-risk statement should call out that we go to CR in this charter period and that some profile features may get scrapped if they are not backed by two implementations.

sebastiankb:
it seems that those assertions was introduced by this PR: https://github.com/w3c/wot-architecture/pull/790

mmccool:
Closing without merging, as agreed, we will take a different approach to solving this problem, which will need a new PR.

mmccool:
(I just deleted the branch as well so I don't accidentally merge this change in the future, given the name of the branch.  I'll start a new branch for future IR updates).

PR #873, opened 2022-11-03
Remove Section 7 assertions

mmccool:
As agreed, this PR downgrades sections 7.4 and 7.5 to informative sections, removing any assertions.  This allows the entirely of section 7 to be informative as previously decided.  It also makes a few other necessary changes consistent with this goal.  Details:
- Small change to the abstract that outlines the sections, which previously said both section 6 and section 7 were normative.   Now it only says section 6 is normative.  
     - NOTE: a section outline does NOT belong in the abstract, for the same reason we can't have references in the abstract: the abstract needs to be self-contained.  The section outline in particular has hyperlinks which would be broken if the abstract is used out of content.
     - I would prefer we delete this list, or move it to the intro (although the intro ALSO has an outline, so it may be redundant.)  As this is an editorial change, however, we can do it post-CR.  
     - I have created an issue (and a PR) so we can follow up: https://github.com/w3c/wot-architecture/issues/874 and https://github.com/w3c/wot-architecture/pull/876
- Downgrade of normative references to WOT-PROFILES and WOT-DISCOVERY to informative references (this is consistent with other building block references in section 7 and is also necessary to avoid ReSpec warnings).
- Downgrade or remove assertions in section 7.4.  All but the last assertion were downgraded.  The last assertion, which was on the use of the Discovery spec as a whole, was removed (it's optional anyway).  
- Downgrade assertions in section 7.5.  Generally, this only involved decapitalizing key words and removing spans.  However, there was also a normative reference to RFC2046 that was downgraded to an informative reference (can't have an normative reference in an informative section)
- Updated the CR static version in publication/ver11/3-cr/*.html to be consistent (note: you can view Overview.html if you want to see the assertion highlighting, as this version inlines the CSS file)
- Updated the IR (to remove the deleted assertions)

I also did an actual search in the document for any spans in section 7 that we might have overlooked.  I confirmed that there are no assertions in section 7 after the changes above.  I also checked that there are no ReSpec warnings or errors.  I also checked that removal of these assertions does not change the at-risk section in the sotd; it does not, as none of the removed assertions were at risk.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/873.html" title="Last updated on Nov 3, 2022, 1:28 PM UTC (44c4c35)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/873/1c7ba10...mmccool:44c4c35.html" title="Last updated on Nov 3, 2022, 1:28 PM UTC (44c4c35)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3125,7 +3125,7 @@
         "<a>exploration</a>" mechanisms that provide actual access to metadata.
         </p><p>
         The first and simplest <a>exploration</a> mechanism defines how to fetch a <a>TD</a>
-        directly from a simple web resource or from the device itself. This supports ad-hoc peer-to-peer discovery.
+        directly from the device itself. This supports ad-hoc peer-to-peer discovery.
         However, in many circumstances (including but not limited to management of 
         large collections of TDs), an alternative <a>exploration</a> mechanism,
         the <a>WoT Thing Description Directory</a> (<a>TDD</a>) service, 
@@ -3136,7 +3136,7 @@
         <a>TDDs</a> also provide change notification mechanisms 
         to support secure distribution of
         <a>TD</a> updates to authorized consumers.
-        To preserve security and privacy, any <a>WoT Discovery</a> "exploration"
+        Any <a>WoT Discovery</a> "exploration"
         implementation must only provide metadata and <a>TDs</a>
         after appropriate best-practice authentication and authorization.
         Suitable best-practice security mechanisms for authentication and authorization for 
--- a/index.html
+++ b/index.html
@@ -3102,17 +3102,15 @@
         </p>
         <ul>
         <li>
-        <span class="rfc2119-assertion" id="arch-discovery-no-metadata-in-introduction">
-        Any <a>introduction</a> mechanism used for <a>WoT Discovery</a> MUST NOT 
+        Any <a>introduction</a> mechanism used for <a>WoT Discovery</a> must not
         provide metadata but instead should simply result in 
         one or more opaque URLs at which metadata
         may be accessed, if and only if the user can authenticate and 
-        has the appropriate authorization (access rights).</span>
+        has the appropriate authorization (access rights).
         </li>
         <li>
-        <span class="rfc2119-assertion" id="arch-discovery-no-metadata-in-introduction-urls">
-        <a>Introduction</a> URLs MUST NOT themselves include any metadata,
-        e.g. a human-readable device type or name.</span>  
+        <a>Introduction</a> URLs must not themselves include any metadata,
+        e.g. a human-readable device type or name.  
         Randomly generated URLs are recommended.
         </li>
         </ul>
@@ -3126,7 +3124,7 @@
         "<a>exploration</a>" mechanisms that provide actual access to metadata.
         </p><p>
         The first and simplest <a>exploration</a> mechanism defines how to fetch a <a>TD</a>
-        directly from the device itself. This supports ad-hoc peer-to-peer discovery.
+        directly from a simple web resource or from the device itself. This supports ad-hoc peer-to-peer discovery.
         However, in many circumstances (including but not limited to management of 
         large collections of TDs), an alternative <a>exploration</a> mechanism,
         the <a>WoT Thing Description Directory</a> (<a>TDD</a>) service, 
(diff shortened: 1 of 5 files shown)


mmccool:
Note: I WAS going to add a line "Make section 7 informative" to the change log but it seems we do not have one yet for the changes from the last WD publication.  Since we need a full change log for the CR transition anyway, I will put one together in a separate PR and include that line.

mmccool:
Also the lack of assertion highlighting is due to github.io not picking up testing/atrisk.css, which is linked, not included.  To avoid this problem in the static version I inlined the CSS file.  If you need to see it check out the entire repo and render it locally.  You also won't see it if you just download index.html and render it locally, unfortunately.  If someone can suggest a different approach (e.g. a way to let github.io know it should checkout out specific resources) let me know...

index.html
```diff
@@ -3126,7 +3125,7 @@ <h3>Exploration Mechanisms</h3>
         "<a>exploration</a>" mechanisms that provide actual access to metadata.
         </p><p>
         The first and simplest <a>exploration</a> mechanism defines how to fetch a <a>TD</a>
-        directly from the device itself. This supports ad-hoc peer-to-peer discovery.
+        directly from a simple web resource or from the device itself. This supports ad-hoc peer-to-peer discovery.
```
mmccool, line 3128: note: this change is not directly related to the "make section informative" but is an editorial correction.

index.html
```diff
@@ -3137,10 +3136,9 @@ <h3>Exploration Mechanisms</h3>
         <a>TDDs</a> also provide change notification mechanisms 
         to support secure distribution of
         <a>TD</a> updates to authorized consumers.
-        <span class="rfc2119-assertion" id="arch-discovery-metadata-after-authorization">
-        Any <a>WoT Discovery</a> "exploration"
-        implementation MUST only provide metadata and <a>TDs</a>
-        after appropriate best-practice authentication and authorization.</span>
+        To preserve security and privacy, any <a>WoT Discovery</a> "exploration"
```
mmccool, line 3139: "To preserve security and privacy" it an editorial change; it was added to make the text flow better.

index.html
```diff
@@ -3165,12 +3163,9 @@ <h3>Exploration Mechanisms</h3>
         </p>
        </section>
         <p>
-        The <a>Discovery</a> mechanisms defined in [[!WOT-DISCOVERY]] are not mandatory, but
+        The <a>Discovery</a> mechanisms defined in [[?WOT-DISCOVERY]] are not mandatory, but
         they are designed to meet the above requirements and their use will enhance 
         interoperability and usability.
-        <span class="rfc2119-assertion" id="arch-discovery-should-use-standard">Whenever
```
mmccool, line 3171: note that this assertion was removed.  The idea that use of WoT discovery is optional is covered by the previous sentence.

index.html
```diff
@@ -3243,11 +3238,9 @@ <h3>WoT Binding Templates</h3>
             <li>
                 <b>Subprotocol</b>
                 <p>
-                  <span class="rfc2119-assertion" id="arch-binding-templates-known-subprotocol">
-                    <a>Transport protocols</a> MAY have to use information provided by an extension mechanisms to interact successfully.</span>
-                  <span class="rfc2119-assertion" id="arch-binding-templates-subprotocol-explicit-declaration">
-                  Such <a>subprotocols</a> cannot be identified from the URI scheme alone and MUST be declared explicitly
-                  using the <code>subprotocol</code> field inside forms.</span>
+                  <a>Transport protocols</a> may have to use information provided by an extension mechanisms to interact successfully.
```
mmccool, line 3241: note that grammatically, this was a different use of may anyway and not really an assertion

index.html
mmccool: I decided to revert this change to not trigger discussion about "unrelated changes", although I'll probably want to bring it back in later during the CR period as an editorial change.

mmccool:
Ready for review.

sebastiankb (approved):

egekorkan (approved):

mlagally (approved):

mlagally:
Arch call on Nov 9th.: Agree to merge

PR #883, opened 2022-11-25
Update Implementation Report (add indices, assertions to template)

mmccool:
See also https://github.com/w3c/wot-thing-description/pull/1752

This PR adds indices to the assertions listed in the report, ordered by appearance in the text.  The template.csv includes these indices and also the assertion text.

The tooling has also been improved to deal with child assertions, table assertions, and default assertions which however are not used in this spec, but are used in the TD spec.

Note that I start indices with 1 to be consistent with section numbers and child assertion numbering.

The diffs for template.csv as well as atrisk.csv and atrisk.css are largish due to the reordering and new columns but have the same set of assertions in them as before.


mlagally (approved):
This is a big improvement and adds significant transparency to the implementation report.

egekorkan (approved):

PR #884, opened 2022-11-30
Testing CSV cleanup

mmccool:
- make manual.csv a copy of template.csv (since everything is currently manual) 
- note that this CSV file is referenced from wot-testing for the Dec 2022 testfest


mlagally:
Arch call on Dec 8th:
Agree to merge

PR #886, opened 2022-12-05
Revise (D)TLS-1-2 assertions

### LINKED ISSUE #885: TLS/DTLS Version Assertions Ambiguous
The following assertions are a bit ambiguous: arch-security-consideration-tls-1-3, arch-security-consideration-tls-1-2, arch-security-consideration-dtls-1-3, arch-security-consideration-dtls-1-2
- In particular, the (d)tls-1-3 assertions are currently SHOULD for versions 1.3, then others are MAY for 1.2
- This is more confusing than necessary since what we really wanted was to say use 1.2 at least
- Also, the DTLS-1.3 assertion is at risk due to lack of libraries for it (it is a recent spec release).  But if this assertion does not make it into the spec, the related MAY assertion for DTLS-1.2 alone makes it sound like it is not that strongly recommended, which was not the intention.

The Security TF talked about this and would like to make the following changes to both 1-2 assertions, which are editorial in nature and do not change the intent or implementation requirements (and we already have test cases):

Current text: 
- If TLS 1.3 cannot be used for compatibility reasons but secure transport over TCP is appropriate, TLS 1.2 [RFC5246] MAY be used.
- If DTLS 1.3 cannot be used for compatibility reasons but secure transport over UDP is appropriate, then DTLS 1.2 [RFC6347] MAY be used.

Proposed Revision:
- If TLS 1.3 cannot be used for compatibility reasons but secure transport over TCP is appropriate, then at least TLS 1.2 [RFC5246] <strike>MUST</strike> SHOULD be used.
- If DTLS 1.3 cannot be used for compatibility reasons but secure transport over UDP is appropriate, then at least DTLS 1.2 [RFC6347] <strike>MUST</strike> SHOULD be used.

To Discuss: should we use MUST or SHOULD in the second set of assertions, consistent with the other assertions?  In an ideal world we would insist on (D)TLS but I know there are situations when they are not possible or inconvenient to use, e.g. on a LAN).  See below.  Currently to not change the original intent (and keeping the change editorial) I am intending to use SHOULD.  The original intent of the "MAY" was to *allow* downgrading to 1.2, with the 1.3 assertions taking priority.  But this breaks if the DTLS-1.3 assertion gets removed.

mmccool:
Resolves #885

<strike>Since we would like this to go into the CR version, this PR also updates the ver11/3-cr/*.html files correspondingly.</strike>

Update: We agreed that this is not a good time to disrupt the CR transition process, will attempt to make this (editorial) change during CR period, prior to PR transition.  So this pull request only changes the root index file.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-architecture/pull/886.html" title="Last updated on Apr 6, 2023, 11:12 AM UTC (494c4b1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-architecture/886/49a6909...mmccool:494c4b1.html" title="Last updated on Apr 6, 2023, 11:12 AM UTC (494c4b1)">Diff</a>


JKRhb (approved):

j1y3p4rk (approved):

mmccool:
I have updated this PR to only modify the root index.html file, not the CR snapshot, as discussed.

mmccool:
We should discuss the process for making this change.  Do we need additional security reviews?  Special permission from the director?

mmccool:
Notes:
- There is also already an assertion "_Versions of DLTS or TLS earlier than 1.2 MUST NOT be used for new development._" which actually logically covers the "at least" part of the revised assertion.   
- Since all implementations use at least TLS or DTLS 1.2, however, this proposed change would not impact any implementations.
- Security TF Discussed on Jan 23 and came to the following conclusions:
      - Not worth insisting on another CR to include this change, as logically the meaning is captured considering the above additional assertion
      - Certainly we could include it IF there was another CR for other reasons
      - Do we ask for an exception?  We can ask PLH.
      - Worst case, we don't change it.  While not ideal, this is acceptable.

mmccool:
- check that DLTS is just a typo in my quote, not the spec
- check that new wording does not disallow earlier versions of (D)TLS for describing existing implementations (the "new development" part)

mlagally:
Arch call on April6th: Agree to merge after resolving merge conflict. This must not modify the published CR version.

mmccool:
OK, it turns out there were not only no conflicts, but these changes had already been merged in another PR, so this PR was just a leftover.  I did use it to fix the typo, however (DLTS -> DTLS).

PR #887, opened 2022-12-13
Implementation Report Update - Dec 2022

mmccool:
Include new results from Dec 2022 Testfest.


mmccool:
As Kaz noted, this incorrectly changed the DTLS assertions, need to revert before merging.

ashimura:
FYI, [rendered HTML by Statically](https://cdn.statically.io/gh/w3c/wot-architecture/08fb0e8906559e013b90ab12dd19ad8130504c53/testing/report11.html)

mmccool:
OK, the DTLS thing has been reverted.

PR #920, opened 2023-05-30
Update Implementation Report

mmccool:
- update template, fixing various small issues, making consistent with TD and Discovery
- update language so it relates to PR, not CR (CR period becomes past tense, etc)
- update implementation descriptions
- update copyright and acknowledgements
- no changes to assertion status


mmccool:
Rendered version of the report should be at https://w3c.github.io/wot-architecture/testing/report11.html

PR #926, opened 2023-07-05
Create README.md

ashimura:


publication/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Checklist
+
+* Make sure that you update the recent specification changes
+* For Proposed Recommendation Transition, create an issue at X.
+
+## Files
+
+* Overview.html - static HTML for publication
+* diff.html - diff between the [previous published verion](https://www.w3.org/TR/2023/CR-wot-architecture11-20230119/) and Overview.html
```
danielpeintner, line 14: Noticed minor typo

verion -> version

danielpeintner (commented):
I think we should have this kind of readme in all our repos

egekorkan:
> I think we should have this kind of readme in all our repos

We have it in TD and Binding Template now

ashimura:
I thought all the Editors had installed this kind of README.md to all the repositories already, but the wot-architecture repository didn't really have it.
That's why I've just added it.

We should check all the repositories to make sure, and add this kind of README.md if some of the repos don't have it.

